PR #6, opened 2017-02-06
Fix typo in or1k success injection

dugoh:
This fixes a typo in the arch_set_success for the or1k architecture.


dugoh:
Wrong submission queue

PR #15, opened 2017-10-19
Fix mips cross-compile problem

pr0v3rbs:
Fixed redefine problem and added MACRO value.

I referenced this [chromium page](https://chromium.googlesource.com/android_ndk/+/r14b/sysroot/usr/include/mipsel-linux-android/asm/sysmips.h) for MACRO value.

Thank you.


ldv-alt:
Could you elaborate what are the issues you are trying to solve with these commits, please?
For example, what is required to reproduce this "redefine problem", and why do you want to add fallback definitions for constants defined in <asm/sysmips.h>?

pr0v3rbs:
I tried cross compile with mips architecture. I met two problems.

1. Function redefinition error.
When MIPS macro was defined, than occurred 'sys_set_thread_area' function redefinition error on [ldt.c:112](https://github.com/strace/strace/blob/master/ldt.c#L112). It will has same problem on [ldt.c:112](https://github.com/strace/strace/blob/master/ldt.c#L112), [ldt.c:122](https://github.com/strace/strace/blob/master/ldt.c#L122) when M68K macro was defined.

2. Undeclared variable error.
There was undeclared variable problems on [sysmips.c:53](https://github.com/strace/strace/blob/master/sysmips.c#L53), [sysmips.c:66](https://github.com/strace/strace/blob/master/sysmips.c#L66), [sysmips.c:70](https://github.com/strace/strace/blob/master/sysmips.c#L70), when SETNAME, MIPS_ATOMIC_SET, MIPS_FIXADE macro wasn't defined. So I found reference source code on [chromium google source](https://chromium.googlesource.com/android_ndk/+/r14b/sysroot/usr/include/mipsel-linux-android/asm/sysmips.h), and set macro value.

Thank you.

ldv-alt:
Sorry but your cross for MIPS is broken:
1. struct user_desc is defined on X86 only, so if you see sys_set_thread_area redefinition on MIPS, then it's not a proper MIPS;
2. these constants are always defined on MIPS by <asm/sysmips.h>, so if they are not defined, then it's not a proper MIPS.

pr0v3rbs:
Ok. I found that my mips-linux-musl cross compile environment has some problem.

I apologize for my mistake.

PR #16, opened 2017-10-23
tests: stop passing invalid ops to futex_op_wake

jirislaby:
Kernel 4.14-rc1 (commit https://github.com/torvalds/linux/commit/30d6e0a4190d) started checking op parameter to
futex_op_wake and strace tests started failing due to that:
```
FAIL: futex
===========

futex(0x7fabd78bcffc, 0x5, 0xfacefeed, 0xb, 0x7fabd78bcffc, 0xa0caffee) = -1: Invalid argument
futex.test: failed test: ../futex failed with code 1
```
0xa0caffee decodes as:
`0x80000000 & 0xa0caffee`: oparg is shift
`0x70000000 & 0xa0caffee`: op is FUTEX_OP_OR
`0x0f000000 & 0xa0caffee`: cmp is FUTEX_OP_CMP_EQ
`0x00fff000 & 0xa0caffee`: oparg is sign-extended 0xcaf = -849
`0x00000fff & 0xa0caffee`: cmparg is sign-extended 0xfee = -18

That means the op tries to do this:
```
	(futex |= (1 << (-849))) == -18
```
which is completely bogus. The new check of op in the code is:
```
	if (encoded_op & (FUTEX_OP_OPARG_SHIFT << 28)) {
	    if (oparg < 0 || oparg > 31)
		    return -EINVAL;
	    oparg = 1 << oparg;
	}
```
which results obviously in the "Invalid argument" errno above.

The kernel will soften the failure to print only a (ratelimited)
message, but the tests should be fixed so that we can enforce the error
in the future again.

### DIFF
--- a/tests/futex.c
+++ b/tests/futex.c
@@ -516,8 +516,8 @@ main(int argc, char *argv[])
 			"FUTEX_OP_CMP_EQ<<24|0", ENOSYS, "ENOSYS" },
 		{ 0x80000000, "FUTEX_OP_OPARG_SHIFT<<28|FUTEX_OP_SET<<28|0<<12|"
 			"FUTEX_OP_CMP_EQ<<24|0" },
-		{ 0xa0caffee, "FUTEX_OP_OPARG_SHIFT<<28|FUTEX_OP_OR<<28|"
-			"0xcaf<<12|FUTEX_OP_CMP_EQ<<24|0xfee" },
+		{ 0xa000ffee, "FUTEX_OP_OPARG_SHIFT<<28|FUTEX_OP_OR<<28|"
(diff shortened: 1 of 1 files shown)


esyr:
Well, you can't prohibit strace's testsuite to call whatever syscalls it wants. And it does so for a good measure—to check whether they are decoded correctly. Of course, it would be better to have some shim instead of real kernel for testing these boundary cases that involve undefined behaviour, but, unfortunately, we're not there yet.

So far, I'd like to propose another approach—let's allow the possibility of having EINVAL as one of possible outcomes of the call. For example: https://github.com/esyr/strace/commit/407c2676aff32677dfe766e17bd6468521b19375 . I haven't tested this one yet, however.

PR #55, opened 2018-03-13
a Ubuntu error when using 'INSTALL' to install

bw98:
My programming environment is UBuntu 14.04 x86_64 GNU/Linux. When I use the '/strace' directory 'INSTALL' to install, the following error occurs:
"configure.ac:45: error: possibly undefined macro: AM_INIT_AUTOMAKE
       If this token and others are legitimate, please use m4_pattern_allow.
       See the Autoconf documentation.
Configure.ac:46: error: possibly undefined macro: AM_MAINTAINER_MODE
Configure.ac:902: error: possibly undefined macro: AM_CONDITIONAL”

I used the command "apt-get install libtool" and it was installed successfully. I recorded this error in the 48th line of "INSTALL" 'Note that if you are using Ubuntu, you need install 'libtool',
  Otherwise there is a error "possibly undefined macro: AM_INIT_AUTOMAKE" etc. ”
This problem is common in many Ubuntu distributions such as Ubuntu 16.04 and Ubuntu 14.04.
I hope this record will enable subsequent Ubuntu users to successfully install strace.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -45,7 +45,9 @@ may remove or edit it.
    The file 'configure.ac' (or 'configure.in') is used to create
 'configure' by a program called 'autoconf'.  You need 'configure.ac' if
 you want to change it or regenerate 'configure' using a newer version of
-'autoconf'.
+'autoconf'. Note that if you are using Ubuntu, you need install 'libtool'
+otherwise there is a error "possibly undefined macro: AM_INIT_AUTOMAKE" 
+etc.
 
    The simplest way to compile this package is:
 


ldv-alt:
No, you don't need GNU libtool to build strace.

ldv-alt:
See also INSTALL-git.md file.

bw98:
Yes,if we install strace by ’bootstrap‘ ，no error exist.We need libtool when we use 'automake' to install, however.

esyr:
May I ask you to specify which part of strace's configuration&build infrastructure requires libtool?

bw98:
I make  tests with libtool and without libtool and find that the error has nothing to do with libtool.But I find other thing below.
I review the code once again in both Ubuntu 16.04 and Ubuntu 14.04, and find if I executive 'autoconf' at the first time,there are some error below:
> configure.ac:45: error: possibly undefined macro: AM_INIT_AUTOMAKE
>       If this token and others are legitimate, please use m4_pattern_allow.
>       See the Autoconf documentation.
> configure.ac:46: error: possibly undefined macro: AM_MAINTAINER_MODE
> configure.ac:921: error: possibly undefined macro: AM_CONDITIONAL
and 'autoconf' once again,there is no error.
So the fact is that the error only occurs at the first time and report 3 possibly undefined macro.

esyr:
Because these are automake macros, and you have to run `aclocal -I m4` first in order to generate `aclocal.m4` properly. And I don't quite understand why you're trying to run `autoconf` directly and not `./bootstrap`, as [INSTALL-git.md](../blob/master/INSTALL-git.md) suggests.

bw98:
thanks for the reply ;-) . That's because I originally planned to use INSTALL to install stare. After that I successfully installed strace using INSTALL-git.md.

PR #68, opened 2018-04-18
Make the build reproducible

lamby:
Whilst working on the [Reproducible Builds effort](https://reproducible-builds.org/), we noticed that strace could not be built reproducibly. This is due to the documentation encodes a timezone-varying date in the manpages.

(This was originally filed in Debian as [#896016](https://bugs.debian.org/896016).)

### DIFF
--- a/file-date-gen
+++ b/file-date-gen
@@ -20,6 +20,11 @@ date=
 
 [ -f "${DATE_FILE}" ] && date="$(cat "${DATE_FILE}")"
 
+# Use SOURCE_DATE_EPOCH if it exists.
+# <https://reproducible-builds.org/specs/source-date-epoch/>
+[ -n "${SOURCE_DATE_EPOCH}" ] ||
+	date="@${SOURCE_DATE_EPOCH}"
+
 [ -n "${date}" ] ||
 	date="$(git log -n 1 --format=format:%cD --no-patch "${FILE}")"
 
@@ -32,4 +37,4 @@ date=
 	exit 1
 }
 
-exec printf "%s" $(date "+${DATE_FORMAT}" -d "${date}")
+exec printf "%s" $(date -u "+${DATE_FORMAT}" -d "${date}")


ldv-alt:
I'm sorry but your proposed commit just breaks the script it was supposed to fix:

`+# Use SOURCE_DATE_EPOCH if it exists.`
`+# <https://reproducible-builds.org/specs/source-date-epoch/>`
`+[ -n "${SOURCE_DATE_EPOCH}" ] ||`
`+       date="@${SOURCE_DATE_EPOCH}"`

Contrary to the comment, it ignores SOURCE_DATE_EPOCH if the latter exists, and it makes the script issue an error and print nothing if SOURCE_DATE_EPOCH does not exist.

Besides that, it was wrong to override DATE_FILE and "git log" with SOURCE_DATE_EPOCH.

esyr:
Because both of them are reproducible sources of the file date and make more sense than an arbitrary SOURCE_DATE_EPOCH.

I think that the only problems you actually had are lack of UTC option in `date` invocations and absence of `LC_DATE` reset, as `.strace.1.in.date` file used for identifying `strace.1.in` file date (the only one for which the script is used so far) is generated during build (and goes into the distribution tarball), and builds from git tree use git commit timestamps as a reference.

lamby:
Ah, I see our disconnect now.. so in Debian we try and rebuild everything and — as we don't have your `.git` tree at that point as we build from tarballs — we don't have a date.

Updated patch attached that prefers the git tree so the release behaviour is unchanged and also doesn't break if `SOURCE_DATE_EPOCH` is undefined.

:)

ldv-alt:
Looks like you build from git snapshots (produced by `git archive --format=tar`) instead of released tarballs (produced by `make-dist` script), otherwise `SOURCE_DATE_EPOCH` wouldn't make any difference.

Anyway, please have a look at the current master branch, I suppose it addresses the issue.

lamby:
(Commented on that commit)

PR #85, opened 2018-11-01
INSTALL: mention bootstrap script

dvyukov:
It's bootstrap that generates xlat and configure.
Without bootstrap nothing works.
Mention that it needs to be run as the first step.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -12,8 +12,8 @@ without warranty of any kind.
 Basic Installation
 ==================
 
-Briefly, the shell command './configure && make && make install' should
-configure, build, and install this package.  The following more-detailed
+Briefly, the shell command './bootstrap && ./configure && make && make install'
+should configure, build, and install this package.  The following more-detailed
 instructions are generic; see the 'README' file for instructions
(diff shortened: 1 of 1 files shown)


haoyouab:
Actually there is a file INSTALL-git.md that mentions bootstrap script. People usually miss it, though.

dvyukov:
Right!
I missed it because based on its size it looked like a forwarding reference to INSTALL, which looks like real instructions. And since I've found what looks like instruction for building from source, I did not expect that there are some other instructions for building from some other source.
What is INSTALL for then?

ldv-alt:
We still support installing strace from release tarballs as well as from the git repository.
Unfortunately, we are not very good at separating documentation relevant for git repository users from the rest of documentation, but we are getting better.

For now, please start with README.md which in turn references INSTALL-git.md file.

dvyukov:
OK

dvyukov:
Maybe it makes sense to merge the 2 docs, but say something like `Step 0. run ./bootstrap (skip if you are using a release tarball and already have configure)` to eliminate the documentation indirection for users.

PR #90, opened 2019-01-07
Enable ccache

thedrow:
This should make the build faster.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: c
 
+cache: ccache
+
 dist: trusty
 
 before_install: ci/install-dependencies.sh


ldv-alt:
Sorry, it doesn't make the build noticeably faster because most of the time is taken by tests.
What is certainly does is that it breaks all tests that use gcc-8.

thedrow:
ccache will only make the next build faster.
It caches compilation objects so it avoids compilation of unchanged code.
It seems that ccache triggers a warning which prevents the tests from being compiled:
https://travis-ci.org/strace/strace/jobs/476505765#L3875
I'm not sure why, it shouldn't do so.

ldv-alt:
I'd rather not enable an optimization that causes known regressions.

PR #99, opened 2019-06-10
strace: introduce -fff for timestamp in filename

btwotch:
this is useful when you strace a lot of processes and the pids
on your system wrap around; then output files will be appended.


ldv-alt:
```git grep -F '"%s.%u"'``` shows that ```after_successful_attach``` is not the only place that deals with output file names.
Maybe it's time to move output file name to struct tcb.

chill-otter (approved):

ashutosh887:
I would like to work on this Issue and resolve the conflicts

PR #136, opened 2020-04-21
Support interval specification in when-expression.

zlobober:
Currently, strace allows syscall tampering only for syscall with 1) fixed invocation index or 2) all invocations starting from n-th (possibly with some step). 

Such two options do not cover a useful scenario of testing retrying/backoffing policy of traced program, which may be done by introducing delays/errors on first n invocations, and then stopping tampering.

This PR adds possibility to invoke strace like the following:

```
strace -e inject=recvfrom:delay_exit=100000:when=3,100 ./bin
```

which results in invocations from third to hundredth (inclusive) being tampered. Formally, syntax now looks like `when=first[,last][+step]`.

This change does not affect the existing behavior.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -7,6 +7,8 @@ Noteworthy changes in release ?.? (????-??-??)
   * Updated lists of BPF_* constants.
   * Added -e trace=%clock option for tracing syscalls reading of modifying
     system clocks.
+  * ":when=expr" subexpression now allows specifying last value of the
+    invocation range.
 
 Noteworthy changes in release 5.6 (2020-04-07)
 ==============================================
--- a/defs.h
+++ b/defs.h
@@ -230,6 +230,8 @@ typedef struct ioctlent {
 	|INJECT_F_DELAY_EXIT	\
 	)
 
+# define INJECT_LAST_INF (uint16_t)-1
+
 struct inject_data {
 	uint8_t flags;		/* 6 of 8 flags are used so far */
 	uint8_t signo;		/* NSIG <= 128 */
@@ -240,6 +242,7 @@ struct inject_data {
 
 struct inject_opts {
 	uint16_t first;
+	uint16_t last;
 	uint16_t step;
 	struct inject_data data;
 };
--- a/filter_qualify.c
+++ b/filter_qualify.c
@@ -417,6 +417,7 @@ qualify_inject_common(const char *const str,
 {
 	struct inject_opts opts = {
 		.first = 1,
+		.last = INJECT_LAST_INF,
 		.step = 1,
 		.data = {
 			.delay_idx = -1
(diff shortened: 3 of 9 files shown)


zlobober:
@esyr, @ldv-alt 
Seems like I'm done with tests. Would you mind checking my PR?

PS. I ended up using strace in such manner after Dmitry's "Modern strace" talk, nice material :)

ldv-alt:
On Wed, Apr 22, 2020 at 05:41:10AM -0700, Max Akhmedov wrote:
> Seems like I'm done with tests. Would you mind checking my PR?

While the implementation is simple enough the feature might be useful indeed.
Could you update strace.1.in and add a line to NEWS, please?

> PS. I ended up using strace in such manner after Dmitry's "Modern strace" talk, nice material :)

You are welcome. :)

zlobober:
> Could you update strace.1.in and add a line to NEWS, please?

Done.

ldv-alt:
On Wed, Apr 22, 2020 at 05:41:10AM -0700, Max Akhmedov wrote:
> @esyr, @ldv-alt 
> Seems like I'm done with tests. Would you mind checking my PR?

Hmm, you treat when=65535 as when=65535+.
I suppose this is unintended since you said that
this change does not affect the existing behavior.

Could you have a look into this, please?

zlobober:
Nice observation, let me think how to workaround in the easiest way...

zlobober:
Well, of course the easiest way would be to restrict all three numbers `first`, `last` and `step` to be from  1..65534. Such change affects the existing behavior, but it keeps the code simpler.

But if you want me to, I can move the `last_present` flag into inject_opts structure, and support specifying 65535.

Which option do you prefer? I'm OK with both of them.

ldv-alt:
On Thu, Apr 23, 2020 at 08:45:21AM -0700, Max Akhmedov wrote:
> Well, of course the easiest way would be to restrict all three numbers `first`, `last` and `step` to be from  1..65534. Such change affects the existing behavior, but it keeps the code simpler.
> 
> But if you want me to, I can move the `last_present` flag into inject_opts structure, and support specifying 65535.
> 
> Which option do you prefer? I'm OK with both of them.

Could you have a look at commit 0451f1, please?

ldv-alt:
On Wed, Apr 22, 2020 at 05:41:10AM -0700, Max Akhmedov wrote:
> @esyr, @ldv-alt 
> Seems like I'm done with tests. Would you mind checking my PR?

As @esyr didn't like the proposed ```when=first[,last][+step]``` syntax,
I changed it to ```when=first[..last][+step]``` which is more intuitive.
What do you think about it (commit 75c38356d)?

zlobober:
> Could you have a look at commit 0451f1, please?

I see, you fixed that by using step = 0 for single-value ranges, like it was before. Looks good.

I'm OK with first[..last][+[step]].

PR #140, opened 2020-07-03
s390x: sthyi: Fix section size requirement and error

frankjaa:
Some old z/VM instances report only 40 bytes for the partition
section, as they don't report the last three struct members.

Let's lower the size check of that section to 40 to accommodate them.

Also we expect a section to be at least x bytes, so we should make
that clear in the error message.

Signed-off-by: Janosch Frank <frankja@linux.ibm.com>


ldv-alt:
Apparently this was applied as commit 0ed6f3258d6809e44fc44b3ddfa6d2e996f423e2, thanks.

PR #190, opened 2021-07-20
tests/pidns.c: skip some namespace tests in chroot

chexum:
ioctl(NS_GET_USERNS) fails when run in a chrooted environment,
which is a common way to build with distributions these days.

Handling EPERM allows these tests to be skipped when the
the tests are run in the same environment.

### DIFF
--- a/tests/pidns.c
+++ b/tests/pidns.c
@@ -180,11 +180,20 @@ check_ns_ioctl(void)
 
 	int userns_fd = ioctl(fd, NS_GET_USERNS);
 	if (userns_fd < 0) {
-		if (errno == ENOTTY)
-			error_msg_and_skip("NS_* ioctl commands are not "
-			                   "supported by the kernel");
-		else
+		switch (errno) {
+		case ENOTTY:
+			error_msg_and_skip(
+				"NS_* ioctl commands are not "
+				"supported by the kernel");
+			break;
+		case EPERM:
+			error_msg_and_skip(
+				"NS_* ioctl commands are not "
+				"available while chrooted");
+			break;
+		default:
 			perror_msg_and_fail("ioctl(NS_GET_USERNS)");
+		}
 	}
 
 	close(userns_fd);


ldv-alt:
I regularily run this test in chrooted environments but haven't seen any problems yet.
I wonder what's the real cause of this issue.

chexum:
Does the kernel, with the chrooted environment, have the namespaces support compiled in?  Ie. CONFIG_NAMESPACES and all CONFIG_*_NS?  Are the tests including a full "make check"?

ldv-alt:
This is how it looks in chrooted environments:
```
$ strace -qq -a24 -esignal=none -eioctl,unshare ./kill--pidns-translation >/dev/null 
ioctl(3, NS_GET_USERNS) = 4
unshare(CLONE_NEWUSER|CLONE_NEWPID) = -1 EPERM (Operation not permitted)
unshare: Operation not permitted
```

chexum:
Yes, that's my point - just like unshare() with some options in pidns_test_init(), these ioctls are supposed to fail with EPERM in chroot.  The unshare() code path later checks and makes the test skip, but the ioctls fail and cause a test suite failure.

The pull request is adjusting the conditions so that the ioctl failure (when expected to fail), will make the test skipped instead of failing.

Perhaps a similar effect would be to move check_ns_ioctl() down so that it's not run at all when unshare() failed?

chexum:
Sorry, what is missing is that the environment is not a purely chroot; but set up by the building procedure, I just guessed it to be chroot, but it's also setting up namespaces to mount proc and others, which causes some other namespace calls to fail.   Whether that's a problem or a limitation, I'm not sure.  I've seen similar issues with chrooted environment, but it's clearly just a part of the problem, not the main cause.

However, I'm not able to produce anything standalone to reproduce the same error, other than installing xbps and attempting to build and test the strace package within that build procedure.

What I'm seeing is:

`$ ./xbps-src chroot
=> xbps-src: updating repositories for host (x86_64)...
[*] Updating repository `https://alpha.de.repo.voidlinux.org/current/x86_64-repodata
..
cd builddir/strace-5.12/
[void-packages/masterdir /]$ cd builddir/strace-5.12/
[void-packages/masterdir strace-5.12]$ /bin/pwd
/builddir/strace-5.12
[void-packages/masterdir strace-5.12]$ ./src/strace -qq -a24 -esignal=none -eioctl,unsh>
ioctl(3, NS_GET_USERNS) = -1 EPERM (Operation not permitted)
pidns.c:187: ioctl(NS_GET_USERNS): Operation not permitted
`

PR #196, opened 2021-10-06
tests/pc: avoid program from creating a coredump

rmetrich:
Currently the `pc` test creates a coredump, which pollutes the `/var/lib/systemd/coredump` directory.

This patch proposes to avoid creating the coredump by changing the limit to 0.
Still `systemd-coredump` detects the SEGFAULT and records it in the journal, but at least there is no coredump effectively created.


ldv-alt:
Implemented in commit f78d8e0c2f6155ac7b058269bf4adf04ad473a3c.

PR #200, opened 2021-10-22
Implement futex_waitv syscall decoding

andrealmeid:
Hi,

This is an initial implementation of the new `futex_waitv()` syscall decoding that will be merged at Linux 5.16.

Documentation from the [commit message](https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?id=bf69bad38cf63d980e8a603f8d1bd1f85b5ed3d9):

Add support to wait on multiple futexes. This is the interface implemented by this syscall:

```
futex_waitv(struct futex_waitv *waiters, unsigned int nr_futexes,
	    unsigned int flags, struct timespec *timeout, clockid_t clockid)

struct futex_waitv {
	__u64 val;
	__u64 uaddr;
	__u32 flags;
	__u32 __reserved;
};
```

Given an array of `struct futex_waitv`, wait on each `uaddr`. The thread wakes if a `futex_wake()` is performed at any `uaddr`. The syscall returns immediately if any waiter has `*uaddr != val`. `*timeout` is an optional absolute timeout value for the operation. This syscall supports only 64bit sized timeout structs. The flags argument of the syscall should be empty, but it can be used for future extensions. Flags for shared futexes, sizes, etc. should be used on the individual flags of each waiter.

__reserved is used for explicit padding and should be 0, but it might be used for future extensions. If the userspace uses 32-bit pointers, it should make sure to explicitly cast it when assigning to `waitv::uaddr`.

Returns the array index of one of the woken futexes. There’s no given information of how many were woken, or any particular attribute of it (if it’s the first woken, if it is of the smaller index...).

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,7 @@ Noteworthy changes in release ?.?? (????-??-??)
 * Improvements
   * Implemented --secontext=mismatch option to find mismatches in SELinux
     contexts.
+  * Implemented decoding of futex_waitv syscall introduced in Linux 5.16.
   * Implemented decoding of BPF_LINK_GET_NEXT_ID and BPF_LINK_GET_FD_BY_ID bpf
     syscall commands.
   * Enhanced decoding of BPF_MAP_CREATE, BPF_PROG_TEST_RUN, and BPF_PROG_LOAD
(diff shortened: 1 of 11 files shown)


ldv-alt:
Hi,

Unfortunately, this patch hasn't been posted to the mailing list and hasn't been reviewed in time.

Apparently, there are no tests for this new decoder, also, it doesn't print the array of waiters passed to the syscall.

PR #210, opened 2022-02-02
printsiginfo: decode si_pkey field

bacher09:
This adds decoding of `si_pkey` field  which is set on `SIGSEGV` in case
of memory access violation on some modern CPUs (these have `pku` flag in `/proc/cpuinfo`).

Here's few sample outputs:

```
pkey_alloc(0, 0)                        = 3
pkey_mprotect(0x7f3f20662000, 4096, PROT_READ|PROT_WRITE, 3) = 0
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_PKUERR, si_addr=0x7f3f20662000, si_pkey=3} ---
```
or another case:
```
mmap(NULL, 4096, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f137374b000
mprotect(0x7f137374b000, 4096, PROT_EXEC) = 0
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_PKUERR, si_addr=0x7f137374b000, si_pkey=1} ---
```

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -10,6 +10,7 @@ Noteworthy changes in release ?.?? (????-??-??)
   * Implemented decoding of IFLA_GRO_MAX_SIZE and TCA_ACT_IN_HW_COUNT netlink
     attributes.
   * Implemented decoding of PR_SET_VMA operation of prctl syscall.
+  * Implemented decoding of siginfo_t.si_pkey field.
   * Updated lists of FAN_*, IORING_*, IOSQE_*, KVM_*, MODULE_INIT_*, TCA_ACT_*,
     and *_MAGIC constants.
 
--- a/src/printsiginfo.c
+++ b/src/printsiginfo.c
@@ -161,10 +161,23 @@ print_si_info(struct tcb *tcp, const siginfo_t *sip)
 			tprint_struct_next();
 			PRINT_FIELD_U(*sip, si_stime);
 			break;
-		case SIGILL: case SIGFPE:
-		case SIGSEGV: case SIGBUS:
+		case SIGILL:
+		case SIGFPE:
+		case SIGBUS:
 			tprint_struct_next();
 			PRINT_FIELD_PTR(*sip, si_addr);
+			break;
+		case SIGSEGV:
+			tprint_struct_next();
+			PRINT_FIELD_PTR(*sip, si_addr);
+#ifdef HAVE_SIGINFO_T_SI_PKEY
+			switch (sip->si_code) {
+			case SEGV_PKUERR:
+				tprint_struct_next();
+				PRINT_FIELD_U(*sip, si_pkey);
+				break;
+			}
+#endif
 			break;
 		case SIGPOLL:
 			switch (sip->si_code) {
--- a/configure.ac
+++ b/configure.ac
@@ -423,7 +423,8 @@ AC_CHECK_MEMBERS([struct utsname.domainname],,, [#include <sys/utsname.h>])
 AC_CHECK_MEMBERS(m4_normalize([
 	siginfo_t.si_syscall,
 	siginfo_t.si_timerid,
-	siginfo_t.si_overrun
+	siginfo_t.si_overrun,
+	siginfo_t.si_pkey
 ]),,, [#include <signal.h>])
 
 AC_CHECK_HEADERS(m4_normalize([
--- a/tests/.gitignore
+++ b/tests/.gitignore
@@ -860,6 +860,7 @@ seccomp_get_action_avail
 seccomp_get_notif_sizes
 seccomp_get_notif_sizes-success
 secontext.am
+segv_accerr
 segv_pkuerr
 select
 select-P
--- a/tests/.gitignore
+++ b/tests/.gitignore
@@ -860,6 +860,7 @@ seccomp_get_action_avail
 seccomp_get_notif_sizes
 seccomp_get_notif_sizes-success
 secontext.am
+segv_pkuerr
 select
 select-P
 semop
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -335,6 +335,7 @@ check_PROGRAMS = $(PURE_EXECUTABLES) \
 	seccomp-filter-v \
 	seccomp-strict \
 	seccomp_get_notif_sizes-success \
+	segv_pkuerr \
 	select-P \
 	set_ptracer_any \
 	set_sigblock \
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -335,6 +335,7 @@ check_PROGRAMS = $(PURE_EXECUTABLES) \
 	seccomp-filter-v \
 	seccomp-strict \
 	seccomp_get_notif_sizes-success \
+	segv_accerr \
 	segv_pkuerr \
 	select-P \
 	set_ptracer_any \
--- a/tests/gen_tests.in
+++ b/tests/gen_tests.in
@@ -840,6 +840,7 @@ seccomp-filter-v	-v -e trace=seccomp
 seccomp_get_action_avail	-e trace=seccomp
 seccomp_get_notif_sizes		-etrace=seccomp -a42
 seccomp_get_notif_sizes-success	-einject=seccomp:retval=42 -etrace=seccomp -a42
+segv_accerr	-qq --trace=none
 segv_pkuerr	-qq --trace=none
 select	-a36
 select-P	-a36 -e trace=select -P /dev/full 9>>/dev/full
--- a/tests/gen_tests.in
+++ b/tests/gen_tests.in
@@ -840,6 +840,7 @@ seccomp-filter-v	-v -e trace=seccomp
 seccomp_get_action_avail	-e trace=seccomp
 seccomp_get_notif_sizes		-etrace=seccomp -a42
 seccomp_get_notif_sizes-success	-einject=seccomp:retval=42 -etrace=seccomp -a42
+segv_pkuerr	-qq --trace=none
 select	-a36
 select-P	-a36 -e trace=select -P /dev/full 9>>/dev/full
 semop	-a32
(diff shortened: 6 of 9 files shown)


ldv-alt:
On Wed, Feb 02, 2022 at 03:03:45PM +0000, codecov[bot] wrote:
> > The diff coverage is `33.33%`.
> | [src/printsiginfo.c](https://codecov.io/gh/strace/strace/pull/210/diff?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=strace#diff-c3JjL3ByaW50c2lnaW5mby5j) | `98.27% <33.33%> (-1.73%)` | :arrow_down: |

I wonder whether the new code could be tested?

NEWS
```diff
@@ -12,6 +12,7 @@ Noteworthy changes in release ?.?? (????-??-??)
   * Implemented decoding of PR_SET_VMA operation of prctl syscall.
   * Updated lists of FAN_*, IORING_*, IOSQE_*, KVM_*, MODULE_INIT_*, TCA_ACT_*,
     and *_MAGIC constants.
+  * Add decoding of si_pkey field in siginfo struct
```
ldv-alt, line 15: Please make sure each sentence ends with a dot.

If you have a look at the NEWS file, you'll see that almost every
"Improvements" section ends with "Updated lists of constants" line followed by
"Updated lists of ioctl commands" line.
Let's keep this tradition, please add your line before rather than after
these lines.

src/printsiginfo.c
```diff
@@ -165,6 +165,12 @@ print_si_info(struct tcb *tcp, const siginfo_t *sip)
 		case SIGSEGV: case SIGBUS:
 			tprint_struct_next();
 			PRINT_FIELD_PTR(*sip, si_addr);
+#ifdef HAVE_SIGINFO_T_SI_PKEY
+			if (sip->si_signo == SIGSEGV && sip->si_code == SEGV_PKUERR) {
+				tprint_struct_next();
+				PRINT_FIELD_D(*sip, si_pkey);
```
ldv-alt, line 171: Apparently, siginfo_t.si_pkey is not the only field that could be printed in case of SIGSEGV or SIGBUS,
so I'd suggest to separate `case SIGSEG` and `case SIGBUS`, and use a `switch` on `sip->si_code` instead of an `if` statement.

ldv-alt (changes requested):
commit message: please change
```
* src/printsiginfo.c: decode si_pkey field.
```
to
```
* src/printsiginfo.c (print_si_info) [HAVE_SIGINFO_T_SI_PKEY]: Decode
si_pkey field.
```

bacher09:
> I wonder whether the new code could be tested?

Behavior is different on different platforms. On platforms that have `pku` and `ospke`  in `/proc/cpuinfo` this code would crash with `SIGSEGV` (and `si_pkey=1`) :

``` 
#define _GNU_SOURCE
#include <stdio.h>
#include <unistd.h>
#include <stdlib.h>
#include <sys/mman.h>
#include <sys/prctl.h>

int main(void) {
   prctl(PR_SET_DUMPABLE, 0);
   int *buf = mmap(NULL, getpagesize(), PROT_EXEC, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);

   if (buf == MAP_FAILED) {
       perror("mmap");
       exit(EXIT_FAILURE);
   }
   if (*buf) {
       asm volatile("":::"memory");
   }
   puts("PKU disabled, so it isn't crashed");
}
```

while on other platforms it will succeed (since these allow read access to `PROT_EXEC` regions).

I can try to make conditional test, that would be executed only of `pku` is present in `/proc/cpuinfo`.

ldv-alt:
On Thu, Feb 03, 2022 at 12:14:15PM -0800, Slava wrote:
> > I wonder whether the new code could be tested?
> 
> Behavior is different on different platforms.

That's OK for such platform-specific code.

>    puts("PKU disabled, so it isn't crashed");

Adding
```
	error_msg_and_skip("PKU seems to be disabled");
```
would make the test executable exit with code 77,
this could be used to skip the whole test.

bacher09:
Hey, wanted to ask what do you think about tests like this:

```#define _GNU_SOURCE
#include <stdio.h>
#include <unistd.h>
#include <stdlib.h>
#include <sys/mman.h>
#include <sys/prctl.h>

int main(void)
{
        prctl(PR_SET_DUMPABLE, 0);
        int *buf = mmap(NULL, getpagesize(), PROT_EXEC,
                                        MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);

        if (buf == MAP_FAILED) {
           perror("mmap");
           exit(EXIT_FAILURE);
        }
        asm volatile("":: "r" (*buf));
        puts("SIGSEGV did not happen");
        return 0;
}
```


and script like this:

```#!/bin/sh
#
# Check -i option.
#
# Copyright (c) 2015-2022 The strace developers.
# All rights reserved.
#
# SPDX-License-Identifier: GPL-2.0-or-later

. "${srcdir=.}/init.sh"

check_prog grep
check_prog sed


if [ ! "/proc/cpuinfo" ]; then
        exit 0
fi

if ! grep -E -e '^flags.*pku' /proc/cpuinfo | grep ospke > /dev/null 2>&1; then
        exit 0
fi

# this test works only if pku is present
set -- "../$NAME"
$STRACE -e trace=none "$@" > "$LOG"  2>&1 |:
addr="$(sed -r -n 's/^--- SIGSEGV \{si_signo=SIGSEGV, si_code=SEGV_PKUERR, si_addr=(0x[[:xdigit:]]+),.*/\1/p' $LOG)" &&
[ -n "$addr" ] || dump_log_and_fail_with
cat > "$EXP" << __EOF__
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_PKUERR, si_addr=${addr}, si_pkey=1} ---
+++ killed by SIGSEGV +++
__EOF__

match_diff "$LOG" "$EXP"
```

I was going to improve it a bit, so it would check cases with different `si_pkey`.

bacher09:
I've pushed my changes into another branch (to avoid overwriting your changes), you can check it [here](https://github.com/bacher09/strace/commit/a76c1d36c8535b5a6c61033def95876d615bd749). If it's fine I can force push into this branch.

bacher09:
Hey, I've replaced:

```
if (*p)
    __asm__ volatile("":::"memory");
```

with 

```
__asm__ volatile("":: "r" (*p));
```

which is IMHO a bit simpler and generates a single `mov` instruction (in x86 assembly), unlike trick with if (which produces 2 instructions).

ldv-alt:
On Sun, Feb 06, 2022 at 07:49:14AM -0800, Slava wrote:
> Hey, I've replaced:
> 
> ```
> if (*p)
>     __asm__ volatile("":::"memory");
> ```
> 
> with 
> 
> ```
> __asm__ volatile("":: "r" (*p));
> ```
> 
> which is IMHO a bit simpler and generates a single `mov` instruction (in x86 assembly), unlike trick with if (which produces 2 instructions).

Thanks.
It's clarity that's more important in the test than the number of instructions.

ldv-alt:
On Sat, Feb 05, 2022 at 02:48:31PM -0800, Slava wrote:
> I've pushed my changes into another branch (to avoid overwriting your changes), you can check it [here](https://github.com/bacher09/strace/commit/a76c1d36c8535b5a6c61033def95876d615bd749). If it's fine I can force push into this branch.

I'm sorry, but github is not the place where we prefer to discuss things, see README.md

ldv-alt:
On Sat, Feb 05, 2022 at 05:10:02PM -0800, Slava wrote:
> Hey, wanted to ask what do you think about tests like this:
> 
> ```#define _GNU_SOURCE
> #include <stdio.h>
> #include <unistd.h>
> #include <stdlib.h>
> #include <sys/mman.h>
> #include <sys/prctl.h>
> 
> int main(void)
> {
>         prctl(PR_SET_DUMPABLE, 0);
>         int *buf = mmap(NULL, getpagesize(), PROT_EXEC,
>                                         MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
> 
>         if (buf == MAP_FAILED) {
>            perror("mmap");
>            exit(EXIT_FAILURE);
>         }
>         asm volatile("":: "r" (*buf));
>         puts("SIGSEGV did not happen");
>         return 0;
> }
> ```
> 
> 
> and script like this:
> 
> ```#!/bin/sh
> #
> # Check -i option.

Really? :)

> #
> # Copyright (c) 2015-2022 The strace developers.
> # All rights reserved.
> #
> # SPDX-License-Identifier: GPL-2.0-or-later
> 
> . "${srcdir=.}/init.sh"
> 
> check_prog grep
> check_prog sed
> 
> 
> if [ ! "/proc/cpuinfo" ]; then
>         exit 0
> fi
> 
> if ! grep -E -e '^flags.*pku' /proc/cpuinfo | grep ospke > /dev/null 2>&1; then
>         exit 0
> fi

No, /proc/cpuinfo is not reliable in this respect, there are architectures
like ppc64 and ppc64le that do not have any pku mentioned in
/proc/cpuinfo at all.

> # this test works only if pku is present
> set -- "../$NAME"
> $STRACE -e trace=none "$@" > "$LOG"  2>&1 |:
> addr="$(sed -r -n 's/^--- SIGSEGV \{si_signo=SIGSEGV, si_code=SEGV_PKUERR, si_addr=(0x[[:xdigit:]]+),.*/\1/p' $LOG)" &&
> [ -n "$addr" ] || dump_log_and_fail_with
> cat > "$EXP" << __EOF__
> --- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_PKUERR, si_addr=${addr}, si_pkey=1} ---
> +++ killed by SIGSEGV +++
> __EOF__
> 
> match_diff "$LOG" "$EXP"
> ```

Yet again, this won't play well with other architectures.  For example,
on ppc64 and ppc64le for some reason it's si_pkey=2 instead of si_pkey=1,
and on mips and riscv it's SEGV_ACCERR instead of SEGV_PKUERR.

I'd rather stick with the version of this test that does all necessary
checks inside the executable.

ldv-alt (approved):

PR #219, opened 2022-08-16
ci(lint): add shell linter - Differential ShellCheck

jamacku:
Differential ShellCheck is a GitHub action that performs differential ShellCheck scans on shell scripts changed via PR and reports results directly in PR.

Since this repository contains some shell scripts I think you might find it useful. Differential ShellCheck is able to produce reports in SARIF format, GitHub understands this format and is able to display it nicely as PR comment and on the `Files Changed` tab, please see below.

image

image

Documentation is available at: [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

### DIFF
--- a/.github/workflows/differential-shellcheck.yml
+++ b/.github/workflows/differential-shellcheck.yml
@@ -1,0 +1,38 @@
+---
+# https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme
+
+name: Differential ShellCheck
+on:
+  push:
+  pull_request:
+    branches:
+      - master
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Repository checkout
+        uses: actions/checkout@v3
+        with:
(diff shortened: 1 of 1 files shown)


ldv-alt:
Jan, thank you for the contribution, I agree it makes sense running shellcheck on shell scripts.
Sometimes we need to run tests offline, what could you recommend with regards to running a shellcheck based test this way?

jamacku:
As far as I know, it's not possible to run Differential ShellCheck locally. It was developed for GitHub in the first place. I have recently worked towards enabling Differential ShellCheck for other services like GitLab and potentially local testing. But I can't promise when this will be done and ready to use.

Currently, the only way how to verify your shell script changes before creating PR is to push the changes into a branch (on GitHub), and Differential ShellCheck runs and reports back to you if any defects were added by your commit.

ldv-alt (approved):

PR #228, opened 2022-10-28
Add structured output

lefessan:
Add options "-B / --structured-output FORMAT" to print the output of strace in a more parsable format. Currently, we provide "json" for JSON format and "ocaml" for OCaml format.

Given the extend of modifications, we don't aim at covering all the features of strace, but only the most commonly used, and keep full textual backward compatibility.

JSON Format definition:

The format can be checked by putting it in a contest "[... null]", as an array of syscalls. All ints are stored in strings to avoid overflows and specific notations.

OCaml Format definition:

terminals:= token_string | token_integerL

struct := "[" [ token_string "," value ';' ]* "]"
array := "[" [ value ";" ]* "]"
value :=
| "NULL"
| "STRING" token_string
| "PARTIAL_STRING" token_string
| "ARRAY" array
| "INDIRECT" array
| "INT" token_integer
| "STRUCT" struct
| "FLAGS" array
| "CALL" "(" token_string "," array ")"

Values can be typechecked by the OCaml compiler using the following ADT: type t =
  | NULL
  | PARTIAL_STRING of string
  | STRING of string
  | INT of int64
  | INDIRECT of t list
  | ARRAY of t list
  | FLAGS of t list
  | STRUCT of ( string * t ) list
  | CALL of string * t list
  | ARG of string * t


lefessan:
I fixed the white-spaces error.

ldv-alt:
Fabrice, thanks for taking this on.  I'd like to ask you to follow the coding style we use in strace, which is similar to the one used in the Linux kernel.  In particular, tabs instead of spaces, `if (` instead of `if( `, and so on.  Also, I've already made most of automatic changes in the master branch, please rebase.  Please turn `TCP_STATE_*` macros into an `enum`, and add a comment about `TCP_STATE_DUMMY`, otherwise it's not clear why `tprint_or` is defined using `TCP_STATE_DUMMY` while `tprint_plus` is defined using `TCP_STATE_DELIM`.  I'll try to do a proper review when the code is updated.

lefessan:
@ldv-alt I tried to fix most indentation problems, and rebased it on the latest version. Is there a tool that I can use to spot such problems automatically ?

ldv-alt:
I'm not aware of such a tool although it might exist.
Fabrice, how do you test these changes?  Could these tests be used in CI?

masatake:
https://github.com/torvalds/linux/blob/master/scripts/checkpatch.pl is the one used in the Linux kernel to check the style.

ldv-alt:
OK, I've cherry picked those parts of the PR that could be applied without structured output itself.
@lefessan could you rebase, please, so I'd be able to merge it somehow?

Also, it's crucial to find a way how to test structured output in CI.

ldv-alt:
@lefessan, are you still interested in getting this PR merged?

pombredanne:
@lefessan gentle ping :)

spoutn1k:
Is this abandoned ? I am very interested in JSON output and wanted to know if this feature was desired still and if I should try and merge the changes.

ldv-alt:
Apparently, this PR is stalled for almost a year now.

spoutn1k:
It is stalled, but my question was more about if I should try to finish the work or if the project took another direction and it was not relevant anymore. If I try and merge it, will it even be considered ?

ldv-alt:
The topic is by no means less relevant than before, but it's still non-trivial.

PR #230, opened 2022-12-16
Replace off64_t with off_t and add largefile support knob

kraj:


ldv-alt:
This change breaks at least mpers tests because the same config.h is used to compile both native and -m32 tests.

ldv-alt (changes requested):
I'm not sure it's the right approach, but anyway, the test suite should pass.

PR #236, opened 2023-02-09
Add some kfd ioctl decoding

ZenithalHourlyRate:
kfd (/dev/kfd) is an interface exposed by amdgpu kernel driver for user space drivers (e.g. ROCm)
 
### FIXME 

KD and KFD both use the 'K' base. Fortunately they currently do not overlap and can be distinguished from NR. However, as KFD is gradually adding ioctls, one day they may overlap.

More generously, how do we decode ioctls arg with the same base? We already have `ioctl_next_match` for displaying ioctls with the same code, do we have similar mechanisms?

### Note

TODOs in code will not be addressed in this patchset. They could be implemented later when I get the time.

Also, the variable length array handling for `struct kfd_ioctl_svm_attribute attrs[];` could be generalized to e.g. `defs.h` then `print_field.h`. I think `tfetch_obj` after the end of the struct is quite unsafe. If kernel returned an invalid `nattr`, e.g. -1, then we will access the whole memory. What's your opinion on this.

### TODO

* The FIXME mentioned above
* Tests


src/ioctl.c
```diff
@@ -371,7 +372,12 @@ ioctl_decode(struct tcb *tcp)
 	case 'I':
 		return inotify_ioctl(tcp, code, arg);
 	case 'K':
-		return kd_ioctl(tcp, code, arg);
+		/* FIXME: this is ugly, and wont work forever */
+		if (_IOC_NR(code) >= _IOC_NR(KIOCSOUND)) {
```
masatake, line 376: Though I didn't finish revising after reviewing, the patch I proposed in https://lists.strace.io/pipermail/strace-devel/2022-March/010996.html adds the ability to distinguish the type of file descriptors to strace. Just comment.

ldv-alt:
On Thu, Feb 09, 2023 at 08:21:08PM -0800, Masatake YAMATO wrote:
> Though I didn't finish revising after reviewing, the patch I proposed in https://lists.strace.io/pipermail/strace-devel/2022-March/010996.html adds the ability to distinguish the type of file descriptors to strace. Just comment.

By the way, do you have any plans revisiting that series?

masatake:
> By the way, do you have any plans revisiting that series?

Yes, I'm revising now. I'm also interested in kfd. So I would like to make my patches working well with this pull request.

masatake:
@ZenithalHourlyRate @ldv-alt 
My original patchset for resolving ioctl command overwrapping comes from two parts:
A. getting file descriptor-related information from printdev(),
B. use the information to resolve the command overwrapping between TTY and SND devices.

However, the development of the patchset was staneeded morenot find enough timestrafe'srstand strace's printing routines and TTY layers of Linux.

So I will split the patchset into two, and I will focus on the A part.
However, submitting A. without a real use case is a bad idea.
So I'm interested in this pull request.

Could you see https://github.com/masatake/strace/commits/finfo-v0 especially
https://github.com/masatake/strace/commit/58922f6dcbacdd60afd6d493cf27dfce63e848e5
?

ldv-alt:
The main problem with this PR is the absence of tests for the new ioctl decoder.

ldv-alt:
> @ZenithalHourlyRate @ldv-alt My original patchset for resolving ioctl command overwrapping comes from two parts: A. getting file descriptor-related information from printdev(), B. use the information to resolve the command overwrapping between TTY and SND devices.
> 
> However, the development of the patchset was staneeded morenot find enough timestrafe'srstand strace's printing routines and TTY layers of Linux.
> 
> So I will split the patchset into two, and I will focus on the A part. However, submitting A. without a real use case is a bad idea. So I'm interested in this pull request.

To be honest, I think your original patchset was closer to be merged because, unlike this PR, it contained tests.

ZenithalHourlyRate:
> unlike this PR, it contained tests.

I have added tests now. The data in the tests are collected from my GPU program and I used a script to generate the test functions.

ioctls that my GPU program does not call are now deleted and wont be addressed in this patchset due to a limited time budget. They have beed splitted to [kfd-dev branch](https://github.com/ZenithalHourlyRate/strace/tree/kfd-dev) and anyone interested could add tests for them and get them upstreamed.

> Could you see https://github.com/masatake/strace/commits/finfo-v0 especially
> [masatake@58922f6](https://github.com/masatake/strace/commit/58922f6dcbacdd60afd6d493cf27dfce63e848e5)

I've tested this branch on my machine and it worked.

I am marking this PR as ready for review now. The FIXME mentioned above could be addressed if masatake's patchset get merged first, but I'll leave it for now.

masatake:
> I have added tests now. The data in the tests are collected from my GPU program and I used a script to generate the test functions.

Can I see the script?
I found /dev/kfd in my machine but I have no knowledge about it. So your script helps me.

> ioctls that my GPU program does not call are now deleted and wont be addressed in this patchset due to a limited time budget. They have beed splitted to [kfd-dev branch](https://github.com/ZenithalHourlyRate/strace/tree/kfd-dev) and anyone interested could add tests for them and get them upstreamed.

I would like to try to push your efforts to the upstream.

The parts of my patches were merged into the master branch.
However, the patches are a subset of what the kfd ioctl decoder needs.
Linux kernel assigns the major number of kfd dynamically. The merged patches provide only major numbers to ioctl decoders.

These two patches are needed:

https://github.com/masatake/strace/commit/2aa88bd278e0baee316045ab168d7157b8a4c144
https://github.com/masatake/strace/commit/6bd09fea06f6369fe1c5f0d0a6431201f747d647

@ldv-alt Should I use strace-list to get your review? or can I use this GitHub interface?

ZenithalHourlyRate:
> Can I see the script?

https://gist.github.com/ZenithalHourlyRate/3eb418e2ab231c81ca8a7975b6dad07d

masatake:
@ZenithalHourlyRate Thanks.

ldv-alt:
> These two patches are needed:
> 
> [masatake@2aa88bd](https://github.com/masatake/strace/commit/2aa88bd278e0baee316045ab168d7157b8a4c144) [masatake@6bd09fe](https://github.com/masatake/strace/commit/6bd09fea06f6369fe1c5f0d0a6431201f747d647)
> 
> @ldv-alt Should I use strace-list to get your review? or can I use this GitHub interface?

I prefer the mailing list, but I can cope with the github interface, too.

ZenithalHourlyRate:
I've gone through the new finfo interface, there are some questions regarding how it can be used for kfd ioctl decoding

In `ioctl_decode`, `finfo` is NULL when it is leaving or the fd is not valid
  * kfd and kd still needs distinguishing when in leaving stages
  * in tests we are always using fd = -1

The leaving stage problem can be resolved by changing the logic in `SYS_FUNC(ioctl)`, but for the tests there are no way to provide finfo.

In summary, there must be some fallback ways to distinguish kfd and kd, and the code looks like this

```
if (finfo) {
  // call kfd_ioctl or kd_ioctl accoring to the finfo
} else {
  if (_IOC_NR(code) >= _IOC_NR(KIOCSOUND))
  // ...
}
```

And the current ugly hack can not be avoided.

Since we always need a fallback way, and there is actualy no overlap for now, I think kfd/kd distinguish can just use the hacky way for now to reduce the complexity. In the future when there are overlaps, finfo could be used.

masatake:
See the last three commits in https://github.com/masatake/strace/commits/kfd.

ZenithalHourlyRate:
> Even the last commit () my branch uses your original fallback.

@ldv-alt How do you think about it.

PR #237, opened 2023-02-16
Solve overlapped ioctl commands

masatake:
This is the successor patch-set sent in the last year (https://lists.strace.io/pipermail/strace-devel/2022-March/011014.html)

Though I got a review on the mailing list, revising the patch set was too hard for me at that time.

----

```
IOCTL command assignments are overwrapped.
We can observe the overwrapping when running vi under strace:

  ioctl(0, SNDCTL_TMR_START or TCSETS, {B38400 opost -isig -icanon -echo ...}) = 0
----------------------------^^

We can remove the ambiguity if strace knows about the object behind a
given file descriptor. In the above example, whether the fd (0) is
associated with a sound device or a terminal is the hint for removing
the ambiguity.

We don't have to develop new code for getting the hint from scratch;
the code implementing -y option gathers information about fd already.

  ioctl(0</dev/pts/26<char 136:26>>, TCSETS, {B38400 opost -isig -icanon -echo ...}) = 0
----------------------^^^^^^^^^^^

Currently, the information gathered with -y option is used only for
printing.  This patch-set utilizes the information for improving the
quality of decoding.

With the patch-set, strace reports the ioctl as following:

  ioctl(0, TCSETS, {B38400 opost -isig -icanon -echo ...}) = 0
```


masatake:
overwrap should be overlap. I will update the commits.

masatake:
I changed the target branch.

masatake:
All these changes are not in the master branch. So I close this.

PR #248, opened 2023-04-04
Command-line long option --argv0=name overrides argv[0] on execve

jreiser:
Enables easy application of strace to multi-call executables which interpret argv[0], such as busybox or kmod.


esyr:
On Tue, Apr 04, 2023 at 11:55:37AM -0700, John Reiser wrote:
> Enables easy application of strace to multi-call executables
> which interpret argv[0], such as busybox or kmod.

Sounds like a useful feature.

> ---
>  src/strace.c | 8 ++++++++
>  1 file changed, 8 insertions(+)
> 
> diff --git a/src/strace.c b/src/strace.c
> index 7a58bdbd6..4963e4251 100644
> --- a/src/strace.c
> +++ b/src/strace.c
> @@ -183,6 +183,7 @@ static size_t tcb_wait_tab_size;
>  #ifndef HAVE_PROGRAM_INVOCATION_NAME
>  char *program_invocation_name;
>  #endif
> +char *argv0; /* override argv[0] on execve */
> 
>  unsigned os_release; /* generated from uname()'s u.release */
> 
> @@ -1526,6 +1527,8 @@ exec_or_die(void)
>  		  seccomp_filtering ? "enabled" : "disabled");
>  	if (seccomp_filtering)
>  		init_seccomp_filter();
> +	if (argv0 && params->argv)
> +		params->argv[0] = argv0;

I think it is better to be updated in startup_child, right after setting
params_for_tracee.argv.

>  	execve(params->pathname, params->argv, params->env);
>  	perror_msg_and_die("exec");
>  }
> @@ -2272,6 +2275,7 @@ init(int argc, char *argv[])
>  		GETOPT_QUAL_DECODE_FD,
>  		GETOPT_QUAL_DECODE_PID,
>  		GETOPT_QUAL_SECONTEXT,
> +		GETOPT_ARGV0,

I would put it after GETOPT_TIPS.

>  	};
>  	static const struct option longopts[] = {
>  		{ "columns",		required_argument, 0, 'a' },
> @@ -2334,6 +2338,7 @@ init(int argc, char *argv[])
>  		{ "decode-pids",required_argument, 0, GETOPT_QUAL_DECODE_PID },
>  		{ "secontext",	optional_argument, 0, GETOPT_QUAL_SECONTEXT },
> 
> +		{ "argv0",	required_argument, 0, GETOPT_ARGV0 },

Ditto, I'd rather put it after the "tips" option.

>  		{ 0, 0, 0, 0 }
>  	};
> 
> @@ -2591,6 +2596,9 @@ init(int argc, char *argv[])
>  		case GETOPT_QUAL_DECODE_PID:
>  			qualify_decode_pid(optarg);
>  			break;
> +		case GETOPT_ARGV0:
> +			argv0 = optarg;
> +			break;
>  		default:
>  			error_msg_and_help(NULL);
>  			break;

Updates of usage(), the doc/strace.1.in and NEWS seem to be missing,
you might want to use commit v4.22~8, for example, as a possible reference.

esyr:
Also, as noted by the codecov report, a respective test is missing.

ldv-alt (changes requested):
Every new feature must be covered by tests.

jreiser:
I received helpful and encouraging remarks in response to my pull request
for adding a new feature.
However, there were some bugs in the strace documentation, `make check`
failed, and I could not do a "local run" in my own environment of the
workflow for continuous integration.  I became discouraged.

Executive summary [TL;DR]:
Strace project was not friendly to me as a new contributor because of "silly" errors.
1. `make check` fails because `--enable-bundled=yes` is not configured.
2. `--enable-code-coverage` is omitted, yet is required of an accepted patch.
3. Parallel `make` is not mentioned, but saves HOURS of time for `make check`.
4. No `clang-format` style file in source, despite a call to follow coding style.
5. No mention of the need to use a Debian-derived system to run workflow.

Details:
The bugs start at the very beginning.  Both the `INSTALL` file and
https://strace.io/wiki/NewContributorGuide say:
```
   ./bootstrap
   ./configure
   make
   make check
   make install
```
That does not work as of commit 7a289e79bb5f14e91c874a3b5429e072e7b50bbd which was top-of-trunk on master branch on April 1. `make check` fails to compile all the tests:
```
In file included from perf_event_open_unabbrev.c:2:
perf_event_open.c: In function ‘print_event_attr’:
perf_event_open.c:393:21: error: ‘struct perf_event_attr’ has no member named ‘config3’; did you mean ‘config’?
  393 |         val = attr->config3;
      |                     ^~~~~~~
      |                     config
make[4]: *** [Makefile:16167: perf_event_open_unabbrev.o] Error 1
```
On April 2, I ran on a pre-release beta of Fedora 38 running kernel 6.2.9
and associated *.h files.  Later I updated to fedora 6.2.10-300.fc38.x86_64
from April 6, and got the same error.  The strace source for
`perf_event_open_unabbrev.c` assumed an even more-recent kernel version.
I found `--enable-bundled=yes`.  That worked, and should have been mentioned
as required for` ./configure`.  Also, `--enable-code-coverage` should be
mentioned in order to enable guidance about required code coverage.

Then I tried `make check`, and waited over two hours for it to finish.
Later I added  `-j4`, which reduced the time to 46 minutes.
(My machine has 4 CPU, 8 GB RAM, 2 MB L2 cache, SSD; /usr/bin/top showed
no swap space used, maximum 1850 MB RAM used.)  Friendly directions should
mention `-jN`, or explain why it should not be used.

The directions for submitting a patch call out coding style, but the
source contains no actual style file for clang-format.  "[Linux] kernel
style, but with some changes" is not specific enough.

Then I looked for the actual code coverage checks, and more generally
for the workflow for Continuous Integration.  A Debian-derived system
is required (`apt-get`, etc.), but is not mentioned in documentation.

ldv-alt:
> Executive summary [TL;DR]: Strace project was not friendly to me as a new contributor because of "silly" errors.
> 
>     1. `make check` fails because `--enable-bundled=yes` is not configured.

Maybe worth mentioning in the docs, but one usually don't have to use `--enable-bundled=yes` at all, you just happened to be unlucky to bump into an incompatible linux headers change a workaround for which had not been committed yet.

>     2. `--enable-code-coverage` is omitted, yet is required of an accepted patch.

It is mentioned in dist/INSTALL, we use it for coverage builds only.

>     3. Parallel `make` is not mentioned, but saves HOURS of time for `make check`.

This was addressed by commit 5afe61afd082d6fb7de6716bddd52bed9fb988e3, thanks.

>     4. No `clang-format` style file in source, despite a call to follow coding style.

There seems to be no clang users here, so feel free to contribute one.

>     5. No mention of the need to use a Debian-derived system to run workflow.

The CI workflow was written for github/gitlab, we don't use it elsewhere.

PR #255, opened 2023-06-02
Identify netlink by proto when msghdr.msg_name is not set

donaldh:
Some clients, e.g. ovs-dpctl, use sendmsg with a connected socket so the address family cannot be identified by msg_name. Look up the fd proto when msg_name is NULL.


ldv-alt (commented):
Makes sense.  Do you think there could be a test for this change in tests/ directory?

donaldh:
> Makes sense. Do you think there could be a test for this change in tests/ directory?

I'll take a look at adding a test.

donaldh:
Added a sendmsg() test to tests/netlink_generic.c

Issue #250, opened 2023-04-14
Test suite breaks on 32 bit platforms when strace is built with '-D_TIME_BITS=64 -D_FILE_OFFSET_BITS=64'


kanavin:
To make 32 bit glibc based platforms prepared for year 2038, all software must be rebuilt with the above flags, which ensure that glibc defines time-sensitive data types with enough bits to go beyond that date.

Unfortunately, this breaks strace's test suite, which would use these data types but continue to directly issue 32 bit versions of system calls. For example, ppoll (first lines are printed by strace, second lines by tests/ppoll - they should match but they aren't):

```
# strace -s 2 -e ppoll tests/ppoll
ppoll(NULL, 4207869677, NULL, NULL, 3134983661) = -1 EINVAL (Invalid argument)
ppoll(NULL, 4207869677, NULL, NULL, 3134983661) = -1 EINVAL (Invalid argument)
ppoll(0xb7d70000, 42, 0xb7d70008, 0xb7d70010, 8) = -1 EFAULT (Bad address)
ppoll(0xb7d70000, 42, 0xb7d70008, 0xb7d70010, 8) = -1 EFAULT (Bad address)
ppoll(NULL, 0, {tv_sec=-559038737, tv_nsec=0}, NULL, 8) = -1 EINVAL (Invalid argument)
ppoll(NULL, 0, {tv_sec=3735928559, tv_nsec=4207869677}, NULL, 8) = -1 EINVAL (Invalid argument)
ppoll(NULL, 0, {tv_sec=-559038737, tv_nsec=3405705229}, NULL, 8) = -1 EINVAL (Invalid argument)
ppoll(NULL, 0, {tv_sec=-3819351491602432273, tv_nsec=4207869677}, NULL, 8) = -1 EINVAL (Invalid argument)
ppoll([{fd=3, events=POLLIN|POLLPRI|POLLRDNORM|POLLRDBAND}, {fd=4, events=POLLOUT|POLLWRNORM|POLLWRBAND}, ...], 4, {tv_sec=42, tv_nsec=0}, [USR2 CHLD], 8) = 2 ([{fd=4, revents=POLLOUT|POLLWRNORM}, {fd=6, revent)
[...]

PR #263, opened 2023-08-23
tests: do not use libc definitions for timespec/timeval

kanavin:
libc does not guarantee that they match syscall structures, and indeed, they don't if components are built against glibc with -D_TIME_BITS=64 on 32 bit platforms (to avoid collapse in 2038).

Resolves: https://github.com/strace/strace/issues/250

### DIFF
--- a/tests/aio.c
+++ b/tests/aio.c
@@ -10,6 +10,7 @@
 
 #include "tests.h"
 #include "scno.h"
+#include "kernel_old_timespec.h"
 
 #ifdef __NR_io_getevents
 # include <fcntl.h>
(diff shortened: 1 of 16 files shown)


kanavin:
```
recvmmsg-timeout.c: In function ‘main’:
recvmmsg-timeout.c:39:43: error: passing argument 5 of ‘recv_mmsg’ from incompatible pointer type [-Werror=incompatible-pointer-types]
   39 |         int rc = recv_mmsg(-1, &mh, 1, 0, ts);
      |                                           ^~
      |                                           |
      |                                           kernel_old_timespec_t *
In file included from recvmmsg-timeout.c:11:
tests.h:411:66: note: expected ‘struct timespec *’ but argument is of type ‘kernel_old_timespec_t *’
  411 | int recv_mmsg(int, struct mmsghdr *, unsigned int, unsigned int, struct timespec *);
      |                                                                  ^~~~~~~~~~~~~~~~~
```

Not sure why this didn't occur in local testing. Fixup incoming.

kanavin:
Ok, looks much better! Not sure where the codecov fail comes from or even if it's important.

PR #264, opened 2023-08-23
tests/time.c: print syscall return values as unsigned decimal

kanavin:
strace prints them that way, and so the test has to follow, even if the actual time_t is signed.

### DIFF
--- a/tests/time.c
+++ b/tests/time.c
@@ -26,7 +26,7 @@ main(void)
 	time_t t = syscall(__NR_time, NULL);
 	if ((time_t) -1 == t)
 		perror_msg_and_skip("time");
-	printf("time(NULL) = %lld (", (long long) t);
+	printf("time(NULL) = %llu (", zero_extend_signed_to_ull((long) t));
 	print_time_t_nsec(t, 0, 0);
 	puts(")");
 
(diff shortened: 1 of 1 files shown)


ldv-alt (commented):
In line with your changes in #263, we shouldn't be using `time_t`, neither in the decoder nor in the test of this syscall.  As the kernel uses `__kernel_old_time_t` which is the same as `kernel_long_t`, we should rather use `kernel_long_t`.

There might be a difficulty testing this on architectures where sizeof(long) < sizeof(kernel_long_t) such as x32 because `syscall` function may truncate the return value to fit `long`.  In tests/times.c there is a workaround for a similar problem.

kanavin:
> In line with your changes in #263, we shouldn't be using `time_t`, neither in the decoder nor in the test of this syscall. As the kernel uses `__kernel_old_time_t` which is the same as `kernel_long_t`, we should rather use `kernel_long_t`.
> 
> There might be a difficulty testing this on architectures where sizeof(long) < sizeof(kernel_long_t) such as x32 because `syscall` function may truncate the return value to fit `long`. In tests/times.c there is a workaround for a similar problem.

Should that be a separate ticket and PR? This change is specifically about making the test match the code being tested. I have discovered the issue on a 32 bit system (with 64 bit time_t) after setting time to 2040, when the test starts expecting a negative value, but strace still prints a positive one.

PR #275, opened 2023-11-13
Fix possible null dereference.

chestnykh:
strace.c:
- `filename` pointer has dereferenced many times earlier so the check that filename is not NULL doesn't needed.
- current_tcp pointer must be checked before each dereferencing

### DIFF
--- a/src/strace.c
+++ b/src/strace.c
@@ -1674,7 +1674,7 @@ startup_child(char **argv, char **env)
 		if (!path || !*path)
 			pathname[0] = '\0';
 	}
-	if (filename && !*pathname)
+	if (!*pathname)
 		error_msg_and_die("Cannot find executable '%s'", filename);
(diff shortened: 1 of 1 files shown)


ldv-alt (changes requested):
According to tests, when this change is applied, strace aborts with `free(): invalid pointer` diagnostics.

chestnykh:
Yes, i've realized that the memory allocated in `expand_tcbtab` shouldn't be freed at the end of this function because it is references later

chestnykh:
Hmm, looks like last CI run got not last revision

ldv-alt:
> Hmm, looks like last CI run got not last revision

Do you suppose github is capable of doing that kind of things?

chestnykh:
I don't know
I will close this PR and create new one :)

PR #276, opened 2023-11-16
Fix possible null dereference.

chestnykh:
strace.c:
1) `filename` pointer has dereferenced many times earlier so the check that filename is not NULL doesn't needed.
2) `current_tcp` pointer must be checked before each dereferencing

### DIFF
--- a/src/strace.c
+++ b/src/strace.c
@@ -1674,7 +1674,7 @@ startup_child(char **argv, char **env)
 		if (!path || !*path)
 			pathname[0] = '\0';
 	}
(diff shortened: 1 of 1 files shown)


ldv-alt (changes requested):
The commit message also says that `current_tcp pointer must be checked before each dereferencing`, but the commit neither explains why every dereference must be guarded nor adds a check before each dereference despite of the claim.

ldv-alt:
Is there any appetite for addressing the issues with this PR, or should it rather be closed instead?

PR #277, opened 2023-11-17
strace -f: fix potential deadlock during cleanup

donghaobo:
strace is likely to block at terminate if tracee is forking. on linux, calling vfork will cause the calling thread to 'D' state until the child process calls execve. But the child process is attached by default and will block when the fork system call returns. therefore strace must detach child process first, otherwise it will enter infinit waiting for the 'D' state thread.

the golang test code:
```
package main
import (
    "os/exec"
    "syscall"
)
var rc = make(chan int, 10)
func run() {
    cmd := exec.Command("sh", []string{"-c", "uname -a"}...)
    attr := syscall.SysProcAttr{ Setpgid: true }
    cmd.SysProcAttr = &attr
    cmd.Run()
    <-rc
    return
}
func main() {
    for {
        rc <- 1
        go run()
    }
}
```
use `go run test.go` to start the test code. and `strace -fp <pid>` attach to the test code. 'Ctrl+c' very likely cannot stop the strace.

### DIFF
--- a/src/strace.c
+++ b/src/strace.c
@@ -1221,6 +1221,77 @@ interrupt_or_stop(struct tcb *tcp)
 	return false;
 }
 
+/* Returns true if the tracee can be passed to droptcb. */
+static bool
+detach_interrupted_or_stopped(struct tcb *tcp, int status)
+{
+	if (!WIFSTOPPED(status)) {
+		/*
+		 * Tracee exited or was killed by signal.
+		 * We shouldn't normally reach this place:
+		 * we don't want to consume exit status.
+		 * Consider "strace -p PID" being ^C-ed:
+		 * we want merely to detach from PID.
+		 *
+		 * However, we _can_ end up here if tracee
+		 * was SIGKILLed.
+		 */
+		return true;
+	}
+	unsigned int sig = WSTOPSIG(status);
+	debug_msg("detach wait: event:%d sig:%d",
+		  (unsigned) status >> 16, sig);
+	if (use_seize) {
+		unsigned event = (unsigned)status >> 16;
+		if (event == PTRACE_EVENT_STOP /*&& sig == SIGTRAP*/) {
+			/*
+			 * sig == SIGTRAP: PTRACE_INTERRUPT stop.
+			 * sig == other: process was already stopped
+			 * with this stopping sig (see tests/detach-stopped).
+			 * Looks like re-injecting this sig is not necessary
+			 * in DETACH for the tracee to remain stopped.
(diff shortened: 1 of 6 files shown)


ldv-alt:
Sorry, I failed to understand what's the problem you're trying to address, and the reproducer written in  a cryptic language doesn't help either.

donghaobo:
> Sorry, I failed to understand what's the problem you're trying to address, and the reproducer written in a cryptic language doesn't help either.

I tried writing a C program and it produced the same problem.
use `strace -f ./test` to start tracing the test program. use `Ctrl + c` to stop the strace, it will blocked.
```
#include <sys/types.h>
#include <sys/wait.h>
#include <unistd.h>

int main(int argc, char *argv[])
{
	int status;
	pid_t pid = vfork();
	if (!pid) {
		sleep(10);
		_exit(0);
	} else if (pid > 0) {
		waitpid(pid, &status, 0);
	}
	return 0;
}
```

src/strace.c
```diff
@@ -3067,7 +3070,22 @@ cleanup(int fatal_sig)
...
+	while (cnt) {
+		int status;
+		pid_t pid = waitpid(-1, &status, __WALL);
+		debug_func_msg("looking at pid %u", pid);
+		struct tcb *tcp = pid2tcb(pid);
+		if (!tcp || !tcp->pid) {
+			int error = ptrace(PTRACE_DETACH, pid, 0, 0);
+			debug_func_msg("ptrace untraced: %u, %x %d", pid, status, error);
+			continue;
+		}
```
ldv-alt, line 3086: Invoking PTRACE_DETACH on a process which is not our tracee doesn't look correct.

src/strace.c
```diff
@@ -3067,7 +3070,22 @@ cleanup(int fatal_sig)
...
+		detach(tcp, false);
```
ldv-alt, line 3087: Invoking detach() twice means invoking tkill or PTRACE_INTERRUPT twice which doesn't look correct either.

ldv-alt (changes requested):

src/strace.c
donghaobo: fixed by split `detach` into `detach_or_stop`  `detached_stoped_pid`  `print_droptcp`

src/strace.c
donghaobo: for the golang test process, the grandchild process will get here, which may be caused by invoking`detach` twice. the new commited code is ok and `PTRACE_DETACH` was deleted.

PR #281, opened 2023-11-21
DON'T MERGE: Debug ioctl kvm run auxstr vcpu x32

masatake:
Just for running test cases.


masatake:
umove returns an error.
I observed (in the log) that kvm_ioinfo->mmap_addr in kvm_ioctl_run_attach_auxstr was 0xfffffffff7fcb000. The address makes tracee_addr_is_invalid return TRUE.

kvm_ioinfo->mmap_addr should be 0xf7fcb000.
When kernel prints an entry /proc/$pid/maps, sign extension is done.
mmap_cache_rebuild_if_invalid() must handle this when reading the entries.

I'll make a pull request.

PR #282, opened 2023-11-22
kvm: use unsigned long type for mmap_addr member of vcpu_info

masatake:


src/kvm.c
```diff
@@ -331,11 +331,13 @@ kvm_ioctl_run_attach_auxstr(struct tcb *const tcp,
...
 {
 	static struct kvm_run vcpu_run_struct;
+	kernel_ulong_t mmap_addr;
 
 	if (info->mmap_len < sizeof(vcpu_run_struct))
 		return;
 
-	if (umove(tcp, info->mmap_addr, &vcpu_run_struct) < 0)
+	 mmap_addr = truncate_kulong_to_current_wordsize(info->mmap_addr);
+	if (umove(tcp, >mmap_addr, &vcpu_run_struct) < 0)
```
ldv-alt, line 340: I bet you mean `mmap_addr`, not `>mmap_addr`.

masatake: Thank you.

ldv-alt: By the way, `mmap_addr` field should rather have some unsigned type, otherwise there could be a sign extension issue in this umove() invocation.

ldv-alt: Also, I don't think truncate_kulong_to_current_wordsize is needed here because the argument doesn't have a kernel_ulong_t type.  However, the type of umoven's argument is kernel_ulong_t, so my bet is that the issue will be fixed as soon as the type of mmap_addr field is fixed.

masatake: I see. I will prepend such a commit to this pull request.
I wonder why I chose long for the purpose...

masatake:
Two new errors were reported but they are nothing to do with this PR.
I will convert this draft pull request to a real pull request.

ldv-alt:
Oops, I've already prepared an alternative commit message in d5866774c6ceccccb3fc968000788c2a94f31166 yesterday.
Now let's choose which one to merge, or maybe make a third one?

masatake:
Your one is much better than mine.
I close this one.

PR #285, opened 2024-01-06
Added Lustre filesystem magic codes

spoutn1k:
Hello,

Very simple addition as I noticed the Lustre distributed filesystem magic codes were not present in the defined values.

I also noticed the project uses `EXT2_SUPER_MAGIC` for `0xef53` when from my understanding the value is used for `ext2`, `ext3`, `ext4` and derivatives. I did not touch that but was wondering if it should be renamed to `EXT_SUPER_MAGIC`.


ldv-alt:
> Hello,
> 
> Very simple addition as I noticed the Lustre distributed filesystem magic codes were not present in the defined values.
> 
> I also noticed the project uses `EXT2_SUPER_MAGIC` for `0xef53` when from my understanding the value is used for `ext2`, `ext3`, `ext4` and derivatives. I did not touch that but was wondering if it should be renamed to `EXT_SUPER_MAGIC`.

That's because include/uapi/linux/magic.h defines `EXT2_SUPER_MAGIC`, not `EXT_SUPER_MAGIC`.

spoutn1k:
So a few tests failed but I cannot get a clear idea as to why given the nature of the changes.

In `clang15-x86-kheaders`, `filter_seccomp-exitkill` failed.
In `gcc13-x86_64-unwind` and `clang-x86_64-unwind`, `kill-on-exit.test`, `strace-k.test`, `strace-k-p.test`, `strace-k-demangle.test`.

> That's because include/uapi/linux/magic.h defines EXT2_SUPER_MAGIC, not EXT_SUPER_MAGIC.

It defines `EXT3/4_SUPER_MAGIC` also: https://github.com/torvalds/linux/blob/de927f6c0b07d9e698416c5b287c521b07694cac/include/uapi/linux/magic.h#L24-L27. And in `strace` the values are only here for the user to have an idea of what is being passed, correct ? Would it not make sense to have it display a more general EXT value rather than a specific EXT2 value on all the ext derivatives ?

ldv-alt:
> So a few tests failed but I cannot get a clear idea as to why given the nature of the changes.
> 
> In `clang15-x86-kheaders`, `filter_seccomp-exitkill` failed. In `gcc13-x86_64-unwind` and `clang-x86_64-unwind`, `kill-on-exit.test`, `strace-k.test`, `strace-k-p.test`, `strace-k-demangle.test`.

This seems to be unrelated to the change.  For some unexplained reason, filter_seccomp-exitkill and kill-on-exit tests tend to hang up in github ci from time to time.  strace-k and strace-k-demangle tests didn't fail before, so there might be some subtle change in the environment provided by github ci runners.

> > That's because include/uapi/linux/magic.h defines EXT2_SUPER_MAGIC, not EXT_SUPER_MAGIC.
> 
> It defines `EXT3/4_SUPER_MAGIC` also: https://github.com/torvalds/linux/blob/de927f6c0b07d9e698416c5b287c521b07694cac/include/uapi/linux/magic.h#L24-L27. And in `strace` the values are only here for the user to have an idea of what is being passed, correct ? Would it not make sense to have it display a more general EXT value rather than a specific EXT2 value on all the ext derivatives ?

strace shows those symbolic names that are provided by the header files, so that user can look for them in the headers.
If there are several synonyms like EXT{2,3,4}_SUPER_MAGIC strace generally prefers to show the oldest of them.

ldv-alt:
> Very simple addition as I noticed the Lustre distributed filesystem magic codes were not present in the defined values.

Where do these constants come from?  They are not defined in the Linux kernel tree nowadays, and before the removal there was just LL_SUPER_MAGIC defined to 0x0BD00BD0 somewhere deep in the staging area. glibc doesn't provide them either, but internally uses just LUSTRE_SUPER_MAGIC defined to 0x0BD00BD0.  Given this level of obscurity and contradictions, I'm not sure providing these symbolic constants in strace would be helpful.

spoutn1k:
They are defined in the Lustre filesystem source code. We use it in HPC, and I was doing some testing in a site with two high performance filesystems when I noticed strace recognized IBM's gpfs' `GPFS_SUPER_MAGIC` and not the open source Lustre's magic.

Btw - I just checked and the GPFS magic is not defined in the kernel sources either, but included in `strace`.

PR #288, opened 2024-02-15
v4l2: add VIDIOC_QBUF_TIME32 and friends

w23:
On 32-bit kernels there are older variants of VIDIOC_QBUF, VIDIOC_DQBUF, and VIDIOC_QUERYBUF ioctls, which have _TIME32 suffix. These are for old timeval_t with 32-bit timestamps. Some 32-bit applications still use these.

Implement printing these variants alongside regular 64-bit time ones.

### DIFF
--- a/src/v4l2.c
+++ b/src/v4l2.c
@@ -11,6 +11,7 @@
 #include "defs.h"
 
 #include DEF_MPERS_TYPE(kernel_v4l2_buffer_t)
+#include DEF_MPERS_TYPE(kernel_v4l2_buffer_time32_t)
 #include DEF_MPERS_TYPE(kernel_v4l2_event_t)
 #include DEF_MPERS_TYPE(kernel_v4l2_timeval_t)
 #include DEF_MPERS_TYPE(struct_v4l2_clip)
@@ -544,6 +545,52 @@ print_v4l2_timeval(const MPERS_PTR_ARG(kernel_v4l2_timeval_t *) const arg)
 		print_v4l2_timeval(&((where_).field_));	\
 	} while (0)
 
+static void
+print_v4l2_buffer_contents(struct tcb *const tcp, const unsigned int code,
+			   const kernel_v4l2_buffer_t *b)
+{
+	if (entering(tcp)) {
+		PRINT_FIELD_XVAL(*b, type, v4l2_buf_types,
+				 "V4L2_BUF_TYPE_???");
+		if (code != VIDIOC_DQBUF) {
+			tprint_struct_next();
+			PRINT_FIELD_U(*b, index);
+		}
+
+		/* The rest of the buffer hasn't been filled in yet */
+		return;
+	}
+
+	if (code == VIDIOC_DQBUF) {
+		tprint_struct_next();
+		PRINT_FIELD_U(*b, index);
+	}
+	tprint_struct_next();
+	PRINT_FIELD_XVAL(*b, memory, v4l2_memories,
+			 "V4L2_MEMORY_???");
+
+	if (b->memory == V4L2_MEMORY_MMAP) {
+		tprint_struct_next();
+		PRINT_FIELD_X(*b, m.offset);
+	} else if (b->memory == V4L2_MEMORY_USERPTR) {
+		tprint_struct_next();
+		PRINT_FIELD_PTR(*b, m.userptr);
+	}
+
+	tprint_struct_next();
+	PRINT_FIELD_U(*b, length);
+	tprint_struct_next();
+	PRINT_FIELD_U(*b, bytesused);
+	tprint_struct_next();
+	PRINT_FIELD_V4L2_BUFFER_FLAGS(*b, flags);
+	if (code == VIDIOC_DQBUF) {
+		tprint_struct_next();
+		PRINT_FIELD_V4L2_TIMEVAL(*b, timestamp);
+	}
+	tprint_struct_next();
+	tprint_more_data_follows();
+}
+
 static int
 print_v4l2_buffer(struct tcb *const tcp, const unsigned int code,
 		  const kernel_ulong_t arg)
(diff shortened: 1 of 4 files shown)


w23:
I'm not at all familiar with the strace codebase/conventions, so would appreciate any feedback on this.
Whether it makes sense to do things this way, or if it should be changed in any major fashion.

Thanks!

ldv-alt:
Looks fine, just two remarks:
- we don't use `//` for comments;
- tests should be updated to check decoding of VIDIOC_QUERYBUF_TIME32, VIDIOC_QBUF_TIME32, and VIDIOC_DQBUF_TIME32.

w23:
Is there a standard way to run only a subset of tests?

ldv-alt:
> Is there a standard way to run only a subset of tests?

Since we use a traditional automake based test driver, something as simple as
```
make -k check TESTS='ioctl_v4l2.gen.test ioctl_v4l2-success.gen.test'
```
would do.

> `make check TESTS='*v4l2*'` fails with many messages like:

That kind of globbing grabs unrelated files, try something like
```
make -s check TESTS='ioctl_v4l2*.gen.test'
```
or, if the build is in a subdirectory,
```
make -s check TESTS="$(ls ../../tests/ioctl_v4l2*.gen.test |sed 's/.*\///')"
```

w23:
Thanks, that helped!

I also had to concatenate multiple lines produces by `ls` into a single line separated by spaces, so the final command to run from the project root is: `make -s -k check TESTS="$(ls tests/*v4l2*.gen.test|sed 's/.*\///'|tr '\n' ' ')"`.

Technically, it does fail with:
```
Making check in tests-m32
msg_control.c: In function ‘main’:
msg_control.c:380:9: error: writing 16 bytes into a region of size 4 [-Werror=stringop-overflow=]
  380 |         memcpy(CMSG_DATA(cmsg), &tv, sizeof(tv));
...
```
But it happens after it has ran all the v4l2 tests and generated a report, so it's not a problem.

src/v4l2.c
```diff
@@ -544,6 +544,52 @@ print_v4l2_timeval(const MPERS_PTR_ARG(kernel_v4l2_timeval_t *) const arg)
 		print_v4l2_timeval(&((where_).field_));	\
 	} while (0)
 
+static void
+print_v4l2_buffer_contents(struct tcb *const tcp, const unsigned int code,
+		  const kernel_ulong_t arg, kernel_v4l2_buffer_t b)
```
ldv-alt, line 549: The `arg` argument seems to be unused here, let's drop it.
The `b` argument is a structure, let's pass a pointer instead.

w23: Addressed. Thanks!

w23:
I see the CI errors. They seem to happen in tests-m32 only, which don't even compile on my machine with the error stated in the logs above, even for vanilla strace.

Was able to reproduce in an Ubuntu VM, investigating.

ldv-alt:
> Technically, it does fail with:
> 
> ```
> Making check in tests-m32
> msg_control.c: In function ‘main’:
> msg_control.c:380:9: error: writing 16 bytes into a region of size 4 [-Werror=stringop-overflow=]
>   380 |         memcpy(CMSG_DATA(cmsg), &tv, sizeof(tv));
> ...
> ```
> 
> But it happens after it has ran all the v4l2 tests and generated a report, so it's not a problem.

This warning is incorrect, it looks rather like a compiler error.  Also, I cannot reproduce it with gcc 12 and gcc 13.

w23:
> > Technically, it does fail with:
> > ```
> > Making check in tests-m32
> > msg_control.c: In function ‘main’:
> > msg_control.c:380:9: error: writing 16 bytes into a region of size 4 [-Werror=stringop-overflow=]
> >   380 |         memcpy(CMSG_DATA(cmsg), &tv, sizeof(tv));
> > ...
> > ```    
> >   
> > But it happens after it has ran all the v4l2 tests and generated a report, so it's not a problem.
> 
> This warning is incorrect, it looks rather like a compiler error. Also, I cannot reproduce it with gcc 12 and gcc 13.

Yeah, it seems like a `gcc (Gentoo 13.2.1_p20240113-r1 p12) 13.2.1 20240113` bug, but at this point I'm reluctant to investigate it further -- the stack of why I'm doing any of this is way too deep already :).

Nevertheless, I reproduced and fixed the tests under local Ubuntu VM with CI-like setup. I couldn't test all the compiler-target permutations locally, though, it just takes too much time.

w23:
Is there anything else that should be done about this PR? E.g. merge commit message formatted in a special way, or a manually squashed single commit, etc.

There are some other v4l2-related changes that I'd like to suggest, which are more about expanding printing of various arguments and fields, and which are built upon changes in this PR. They are a bit unrelated thematically, though, so the plan is to open another PR after this one is merged.

ldv-alt:
> Is there anything else that should be done about this PR? E.g. merge commit message formatted in a special way, or a manually squashed single commit, etc.

I don't think the individual fixup commits in this PR are worth keeping, so please squash them into a single commit.
Also, please rebase it without merge commits.

> There are some other v4l2-related changes that I'd like to suggest, which are more about expanding printing of various arguments and fields, and which are built upon changes in this PR. They are a bit unrelated thematically, though, so the plan is to open another PR after this one is merged.

Sounds good.

w23:
> > Is there anything else that should be done about this PR? E.g. merge commit message formatted in a special way, or a manually squashed single commit, etc.
> 
> I don't think the individual fixup commits in this PR are worth keeping, so please squash them into a single commit. Also, please rebase it without merge commits.

Sure thing, I'll make a squashed and rebased commit.

However, if I force-push it into the same branch here, there's a chance GitHub will corrupt this PR unrepairably. I've ran into that issue multiple times before. Just a heads-up, if it happens I'll create a new pristine PR.

w23:
Alright, force-push seems to have worked just fine.

ldv-alt (approved):

Issue #278, opened 2023-11-17
Add option to always show PID


JakobDev:
I'm currently trying to write a script which shows which file is edit by a process and it's childs. You can filter the output by the process. I'm using `strace -fyy -e trace=%file --decode-pids=comm <command>` and parse the output.  But there's a little problem: If a syscall is made by a child process, strace prints `[pid  {pid}<{name}>] {rest of line}`, but if the syscall is made by the process started by strace, this part if missing, what makes this more complex to parse. It would be nice to have a argument to make the output always starts with  `[pid  {pid}<{name}>] {rest of line}`.

esyr:
You can trick `strace` into doing so by setting the `--output` option, `strace -fo /dev/stderr COMMAND`, for example.

JakobDev:
That seems to solve it. Thanks! But I think it would be better to have a official option than to rely on a workaround, so I keep this Issue open.

sheharyaar:
Hi @ldv-alt , I would like to work on this feature, can you please assign me this ?

ldv-alt:
> Hi @ldv-alt , I would like to work on this feature, can you please assign me this ?

Feel free to work on this if you like.

PR #289, opened 2024-03-01
Print pid prefix for all processes in followfork mode

vbalyasnyy:
* src/strace.c: Check for followfork in pid prefix condition.

Resolves: https://github.com/strace/strace/issues/278

### DIFF
--- a/src/strace.c
+++ b/src/strace.c
@@ -870,7 +870,7 @@ printleader(struct tcb *tcp)
 	set_current_tcp(tcp);
 	current_tcp->curcol = 0;
 
-	if (print_pid_pfx || (nprocs > 1 && !outfname)) {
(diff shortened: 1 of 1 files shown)


ldv-alt:
This would be an incompatible change as you can easily see from the result of the CI run.
Note that #278 is a request for a new option instead of a change in the default behavior.

ldv-alt (changes requested):
Please make sure the change doesn't break backwards compatibility.

vbalyasnyy:
I tried to synchronize the output format of process identifiers for the case of output to the console and to a file.
This fix breaks backward compatibility.

It seems to me that the new option looks redundant.
ok, I'll delete this solution.

PR #290, opened 2024-03-06
Make -p options accept multiple delimiters between PIDs

masatake:
The original code accepts only one delimiter between PIDs. This change relaxes the condition.

* src/strace.c (process_opt_p_list): Skip the consecutive delimiters.
* tests/syntax.sh (check_zero): New helper function.
* tests/options-syntax.test: Add code testing this change.

### DIFF
--- a/tests/syntax.sh
+++ b/tests/syntax.sh
@@ -15,6 +15,17 @@ log_sfx()
 	printf "%.128s" "$*" | tr -c '0-9A-Za-z-=,' '_'
 }
 
+check_zero()
+{
+	local sfx
+	sfx="$(log_sfx "$*")"
+
+	$STRACE "$@" 2> "$LOG.$sfx" > /dev/null || {
+		cat "$LOG.$sfx" >&2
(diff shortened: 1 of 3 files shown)


tests/syntax.sh
```diff
@@ -15,6 +15,14 @@ log_sfx()
 	printf "%.128s" "$*" | tr -c '0-9A-Za-z-=,' '_'
 }
 
+check_zero()
+{
+	$STRACE "$@" 2> "$LOG.$sfx" > /dev/null || {
```
ldv-alt, line 20: Looks like `$sfx` is undefined here, see how it's defined e.g. in `check_h`.

ldv-alt (changes requested):

tests/syntax.sh
masatake: Updated:
```
check_zero()
{
	local sfx
	sfx="$(log_sfx "$*")"
```

ldv-alt (approved):

PR #292, opened 2024-03-08
Show the slave device for the fd opening /dev/ptmx when -yy is given

masatake:
An example output:

     read(32</dev/ptmx<char 5:2->/dev/pts/7>>, "\0\"\\0", 4) = 4

* src/defs.h (struct finfo::tty_index): New member. (get_finfo_for_dev): Add two parameters.
* src/ioctl.c (SYS_FUNC(ioctl)): Pass pid and fd to get_finfo_for_dev().
* src/utils.c (scan_fdinfo, scan_fdinfo_fn): Declare earlier. (is_ptmx): New macro.
(get_pts): New function.
(get_finfo_for_dev): Read the tty-index field in /proc/$pid/fdinfo/$fd and store the value to struct finfo::tty_index.
(printenv): Pass pid and fd to get_finfo_for_dev(). Print the slave device for the fd opening /dev/ptmx as the counter endpoint.
* tests/dev-pty-yy.c: New test case.
* tests/gen_tests.in: Add dev-pty-yy.
* tests/pure_executables.list: Add dev-pty-yy.

### DIFF
--- a/src/defs.h
+++ b/src/defs.h
@@ -682,11 +682,14 @@ struct finfo {
 	bool deleted;
 	struct {
 		unsigned int major, minor;
+
+		/* Use only when major == 5, minor == 2 (/dev/ptmx) */
+		int tty_index;
 	} dev;
 };
 
 extern struct finfo *
-get_finfo_for_dev(const char *path, struct finfo *finfo);
+get_finfo_for_dev(pid_t pid, int fd, const char *path, struct finfo *finfo);
 
 extern int
 term_ioctl_decode_command_number(struct tcb *tcp,
--- a/src/util.c
+++ b/src/util.c
@@ -645,6 +645,43 @@ printsocket(struct tcb *tcp, int fd, const char *path)
 	return true;
 }
 
+typedef bool (*scan_fdinfo_fn)(const char *value, void *data);
+
+static bool
+scan_fdinfo(pid_t pid_of_fd, int fd, const char *search_pfx,
+	    size_t search_pfx_len, scan_fdinfo_fn fn, void *data)
+{
+	int proc_pid = 0;
+	translate_pid(NULL, pid_of_fd, PT_TID, &proc_pid);
+	if (!proc_pid)
+		return false;
+
+	char fdi_path[sizeof("/proc/%u/fdinfo/%u") + 2 * sizeof(int) * 3];
+	xsprintf(fdi_path, "/proc/%u/fdinfo/%u", proc_pid, fd);
+
+	FILE *f = fopen_stream(fdi_path, "r");
+	if (!f)
+		return false;
+
+	char *line = NULL;
+	size_t sz = 0;
+	bool result = false;
+
+	while (getline(&line, &sz, f) > 0) {
+		const char *value =
+			str_strip_prefix_len(line, search_pfx, search_pfx_len);
+		if (value != line && fn(value, data)) {
+			result = true;
+			break;
+		}
+	}
+
+	free(line);
+	fclose(f);
+
+	return result;
+}
+
 struct finfo *
 get_finfo_for_dev(const char *path, struct finfo *finfo)
 {
@@ -705,43 +742,6 @@ printdev(struct tcb *tcp, int fd, const char *path, const struct finfo *finfo)
 	}
 
 	return false;
-}
-
-typedef bool (*scan_fdinfo_fn)(const char *value, void *data);
-
-static bool
-scan_fdinfo(pid_t pid_of_fd, int fd, const char *search_pfx,
-	    size_t search_pfx_len, scan_fdinfo_fn fn, void *data)
-{
-	int proc_pid = 0;
-	translate_pid(NULL, pid_of_fd, PT_TID, &proc_pid);
-	if (!proc_pid)
-		return false;
-
-	char fdi_path[sizeof("/proc/%u/fdinfo/%u") + 2 * sizeof(int) * 3];
-	xsprintf(fdi_path, "/proc/%u/fdinfo/%u", proc_pid, fd);
-
-	FILE *f = fopen_stream(fdi_path, "r");
-	if (!f)
-		return false;
-
-	char *line = NULL;
-	size_t sz = 0;
-	bool result = false;
-
-	while (getline(&line, &sz, f) > 0) {
-		const char *value =
-			str_strip_prefix_len(line, search_pfx, search_pfx_len);
-		if (value != line && fn(value, data)) {
-			result = true;
-			break;
-		}
-	}
-
-	free(line);
-	fclose(f);
-
-	return result;
 }
 
 static bool
--- a/src/util.c
+++ b/src/util.c
@@ -682,8 +682,25 @@ scan_fdinfo(pid_t pid_of_fd, int fd, const char *search_pfx,
 	return result;
 }
 
+static bool
+is_ptmx(const struct finfo *finfo)
+{
+	return finfo->type == FINFO_DEV_CHR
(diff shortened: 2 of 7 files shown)


esyr:
I'm kinda wary about having a closing angle bracket in there, what about `</dev/ptmx<char 5:2, slave:/dev/pts/7>>` or, if it's too mouthful, just  `</dev/ptmx<char 5:2 @/dev/pts/7>>` or something like that?

masatake:
`</dev/ptmx<char 5:2, slave:/dev/pts/7>>` looks long.
I update the change to print `</dev/ptmx<char 5:2 @/dev/pts/7>>`.

Though using a closing angle bracket, socket uses the following notation:
```
recvfrom(3<UNIX-STREAM:[3826969->3810268]>, ...
```

Base on the socket notation, we can use a bit similar notation in ptmx: `<dev/ptmx<char 5:2 [->/dev/pts/7]>>`.

tests/dev-pty-yy.c
```diff
@@ -0,0 +1,36 @@
...
+	int ptmx_fd = posix_openpt(O_RDWR);
+	if (ptmx_fd < 0)
+		perror_msg_and_fail("posix_openpt");
+
+	if (unlockpt(ptmx_fd) < 0)
+		perror_msg_and_fail("unlockpt");
+
+	char *pts = ptsname(ptmx_fd);
+	if (pts == NULL)
+		error_msg_and_fail("failed in ptsname()");
```
ldv-alt, line 30: Since all these errors are framework issues,  I suggest skipping the test instead of failing it.

ldv-alt (changes requested):
The new test might issue a `statfs("/dev/pts", ...)` syscall, so when it happens, the following tests do not pass:

regex
trace_personality_statfs_64
trace_statfs
trace_statfs_like

masatake:
Though I did force-pushed my changes to ptmx-pty branch of mine, the test doesn't start.

src/defs.h
```diff
@@ -682,11 +682,14 @@ struct finfo {
 	bool deleted;
 	struct {
 		unsigned int major, minor;
+		union {
+			int tty_index;
+		};
```
ldv-alt, line 687: A union with a single member?  Why?

tests/dev-pty-yy.c
```diff
@@ -0,0 +1,36 @@
...
+#include <unistd.h>
+#include <fcntl.h>
+#include <sys/ioctl.h>
+
+int
+main(void)
+{
+	int ptmx_fd = open("/dev/ptmx", O_RDWR);
+	if (ptmx_fd < 0)
+		perror_msg_and_skip("posix_openpt");
```
ldv-alt, line 23:
```
        static const char dev_ptmx[] = "/dev/ptmx";
        char *real_ptmx = realpath(dev_ptmx, NULL);
        if (real_ptmx == NULL)
                perror_msg_and_skip("realpath: %s", dev_ptmx);

        int ptmx_fd = open(real_ptmx, O_RDWR|O_NOCTTY);
        if (ptmx_fd < 0)
                perror_msg_and_skip("open: %s", real_ptmx);

...

        printf("close(%d<%s<char 5:2 @%s>>) = %d\n",
                       ptmx_fd, real_ptmx, pts, rc);
```

tests/pure_executables.list
```diff
@@ -52,6 +52,7 @@ dev--decode-fds-dev
 dev--decode-fds-none
 dev--decode-fds-path
 dev--decode-fds-socket
+dev-pty-yy
```
ldv-alt, line 55: .gitignore also needs this

src/defs.h
masatake: I assumed we would add more device-specific members in the future.
`union` may be the hint for potential contributors; it tells the person this is where to add a device-specific member.

Should we use `union` when we need it?

ldv-alt: Of course we can use union, I just didn't see the point in using it here, I had no idea you're expecting more device-specific members in the future.

src/util.c
```diff
@@ -645,8 +645,33 @@ printsocket(struct tcb *tcp, int fd, const char *path)
...
+	errno = 0;
+	long i = strtol(value, NULL, 10);
+	if (i < 0 || i > INT_MAX || errno) {
+		debug_func_perror_msg("converting \"%s\" to int", value);
+		return false;
+	}
+
+	finfo->dev.tty_index = i;
+	return true;
+}
```
ldv-alt, line 671: Can't we use something from src/string_to_uint.h here?

masatake: Updated.

src/util.c
```diff
@@ -645,8 +645,26 @@ printsocket(struct tcb *tcp, int fd, const char *path)
 	return true;
 }
 
+typedef bool (*scan_fdinfo_fn)(const char *value, void *data);
+
+static bool
+scan_fdinfo(pid_t pid_of_fd, int fd, const char *search_pfx,
+	    size_t search_pfx_len, scan_fdinfo_fn fn, void *data);
```
ldv-alt, line 652: The last question is whether we could rearrange some code to avoid these forward declarations.

src/util.c
```diff
@@ -645,8 +645,26 @@ printsocket(struct tcb *tcp, int fd, const char *path)
...
+
+#define is_ptmx(FINFO) ((FINFO)->type == FINFO_DEV_CHR \
+			&& (FINFO)->dev.major == 5 \
+			&& (FINFO)->dev.minor == 2)
```
ldv-alt, line 656: I think this could be made a static function.

src/util.c
```diff
@@ -645,8 +645,26 @@ printsocket(struct tcb *tcp, int fd, const char *path)
...
+
+static bool get_pts(const char *value, void *data)
+{
+	struct finfo *finfo = data;
+
+	finfo->dev.tty_index = string_to_uint_ex(value, NULL, INT_MAX, "\n");
+	return true;
+}
```
ldv-alt, line 664: The longer I look at this function, the more it seems to me that `get_pts` is not the right name for it.  What do you think about `set_tty_index`?

tests/dev-pty-yy.c
```diff
@@ -0,0 +1,43 @@
...
+
+int
+main(void)
+{
+	static const char dev_ptmx[] = "/dev/ptmx";
+	char *real_ptmx = realpath(dev_ptmx, NULL);
+	if (real_ptmx == NULL)
+		perror_msg_and_skip("realpath: %s", dev_ptmx);
+
+	int ptmx_fd = open(real_ptmx, O_RDWR);
```
ldv-alt, line 27: Let's open it with `O_RDWR|O_NOCTTY`.

src/util.c
ldv-alt: Rearranged.

tests/dev-pty-yy.c
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+#include "tests.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <sys/ioctl.h>
+#include <limits.h>
```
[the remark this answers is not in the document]
ldv-alt: Removed redundant include.

tests/dev-pty-yy.c
ldv-alt: Changed to that effect.

ldv-alt:
Also replaced realpath(3) in the test with our own get_fd_path() to avoid unwanted side effects.

ldv-alt (approved):

PR #308, opened 2024-07-15
print_utils.h: initialize `hex_chars` as an array, not via string

trofi:
`gcc-15` added a new warning in https://gcc.gnu.org/PR115185:

    print_utils.h:16:35: error: initializer-string for array of 'char' is too long [-Werror=unterminated-string-initialization]
       16 | static const char hex_chars[16] = "0123456789abcdef";
          |                                   ^~~~~~~~~~~~~~~~~~

`strace` does not need to store '\0'. We could either initialize the arrays with individual bytes or allocate extra byte for null.

This change initializes the array bytewise.

### DIFF
--- a/src/print_utils.h
+++ b/src/print_utils.h
@@ -12,8 +12,13 @@
 # include <inttypes.h>
 
 /* Hexadecimal output utils */
-
-static const char hex_chars[16] = "0123456789abcdef";
+static const char hex_chars[16] = {
+	/* Did not use "0123456789abcdef" as an initializer directly as
+	 * gcc-15 warns about the truncation:
+	 *   https://gcc.gnu.org/PR115185 */
+	'0', '1', '2', '3', '4', '5', '6', '7',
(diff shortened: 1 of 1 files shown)


src/print_utils.h
```diff
@@ -13,7 +13,7 @@
 
 /* Hexadecimal output utils */
 
-static const char hex_chars[16] = "0123456789abcdef";
+static const char hex_chars[17] = "0123456789abcdef";
```
ldv-alt, line 16: Is there any way to tell the compiler that `hex_chars` is *not* a C string?
If not, then I'd rather have an ugly `{ '0', '1', ..., 'f' }` initialization instead.
In any case, let's add a comment next to the initialization explaining the source of this ugliness.

trofi: AFAIU there is no nice way to annotate it. Switched to `{ '0', '1', ..., 'f' }` array style initializer and added a comment.

ldv-alt (approved):

PR #333, opened 2025-01-20
tests: Reduce expected precision for relative-timestamps tests

andreabolognani:
Currently we have a very small amount of tolerance around the expected 1.0s sleep time, specifically 0.1s in either direction.

If the machine is not very fast (e.g. StarFive VisionFive 2, a very popular and reasonably performant riscv64 board that's now a couple of years old) and is heavily loaded (e.g. because other builds are running at the same time or the test suite has been executed with 'make -j check'), then the test might fail because of an additional delay that's as small as 0.1s.

Since the test is really intended to ensure that the timestamps are displayed using the requested precision, reporting a failure in this case is not very useful. Instead, we should tolerate delays of up to 1.0s as long as the timestamps are printed out using the expected format.


andreabolognani:
@ldv-alt I spotted this while making sure that the upcoming v6.13 is still in good shape for riscv64. Apologies for not catching it earlier.

It would be amazing if the tweak could be included in the upcoming release, as that would allow us to rebuild the Fedora package completely unpatched on riscv64 once the rebase lands there. It would obviously benefit other distros as well.

andreabolognani:
I believe that the CI failures are because of the current status of Rawhide (GCC 15, mass rebuild and all that) rather than my changes.

ldv-alt:
> I believe that the CI failures are because of the current status of Rawhide (GCC 15, mass rebuild and all that) rather than my changes.

Unfortunately, gcc 15 from Rawhide broke our build right before the release.
For example, -Wstringop-overflow now generates false positives with tests/msg_control.c,
and new option -Wunterminated-string-initialization added to -Wextra produces tones of useless warnings in strace tests.

There is no time to wait for gcc in Rawhide to be fixed, so I had to push a commit that turns off -Werror for these warnings in tests.

ldv-alt:
> One of the codecov jobs is failing, though I don't understand how code coverage could possibly be affected.

There seems to be some fluctuation in the coverage,  ±1 line, let's ignore it for now.

ldv-alt (approved):

andreabolognani:
@ldv-alt thanks a lot!

PR #336, opened 2025-02-12
secontext: fixed retrieval of symlink context and /proc/self/ handling

rmetrich:
There were multiple issues with symlink handling.

1. When the path being queried is a symlink, we expect the context of the symlink to be displayed, not the one of the target. Due to this, lgetfilecon() has to be used instead of getfilecon() in selinux_getfilecon().  For selinux_getfdcon(), this doesn't apply because we cannot query a symlink.

2. When in 'mismatch' mode, querying the expected context for symlinks becomes harder, because realpath() cannot be used, since it would resolve the symlink, hence the query would end up on the target instead of the symlink itself. To solve this, only the dirname of the path is resolved.

3. SELinux context queries are performed through prefixing by /proc/<PID>/{root,cwd,fd/<FD>}/. This led to querying strace's namespace instead of PID's namespace when resolving paths such as /proc/self/something, because /proc/self always points to ourselves (hence strace). Because of this /proc/self must be substituted as /proc/<PID> first, but this may not always work, because there could be a "fake" /proc/self" in the path. To solve this, the context query is performed on the substituted path and if there is no result, the original path is queried as fallback.

4. Unrelated to symlinks, there was no distinction between not being able to retrieve the expected context due to some error and not having any known expected context. To solve this, on error, a "??" static context is returned.

### DIFF
--- a/src/secontext.c
+++ b/src/secontext.c
@@ -9,6 +9,7 @@
 
 #include <stdlib.h>
 #include <fcntl.h>
+#include <libgen.h>
 #include <limits.h>
 #include <sys/stat.h>
 #include <unistd.h>
@@ -20,6 +21,8 @@
 #include "secontext.h"
 #include "xmalloc.h"
 #include "xstring.h"
+
+static char unknown_expected_context[] = "??";
 
 /**
  * @param secontext Pointer to security context string.
@@ -116,10 +119,14 @@ selinux_getfdcon(pid_t pid, int fd, char **secontext, char **expected)
 	if (!proc_pid)
 		return -1;
 
-	char linkpath[sizeof("/proc/%u/fd/%u") + 2 * sizeof(int)*3];
-	xsprintf(linkpath, "/proc/%u/fd/%u", proc_pid, fd);
+	int rc;
+	char buf[PATH_MAX + 1];
 
-	int rc = getfilecon(linkpath, secontext);
+	rc = snprintf(buf, sizeof(buf), "/proc/%u/fd/%u", proc_pid, fd);
+	if ((unsigned int) rc >= sizeof(buf))
+		return -1;
+
+	rc = getfilecon(buf, secontext);
 	if (rc < 0 || !is_number_in_set(SECONTEXT_MISMATCH, secontext_set))
 		return rc;
 
@@ -127,7 +134,6 @@ selinux_getfdcon(pid_t pid, int fd, char **secontext, char **expected)
 	 * We need to resolve the path, because selabel_lookup() doesn't
 	 * resolve anything.
 	 */
-	char buf[PATH_MAX + 1];
 	ssize_t n = get_proc_pid_fd_path(proc_pid, fd, buf, sizeof(buf), NULL);
 	if ((size_t) n >= (sizeof(buf) - 1))
 		return 0;
@@ -137,10 +143,11 @@ selinux_getfdcon(pid_t pid, int fd, char **secontext, char **expected)
 	 * may be reused by a different file with different context.
 	 */
 	strace_stat_t st;
-	if (stat_file(linkpath, &st))
+	if (stat_file(buf, &st) < 0)
 		return 0;
 
-	get_expected_filecontext(buf, expected, st.st_mode);
+	if ((get_expected_filecontext(buf, expected, st.st_mode) < 0) && (errno != ENOENT))
+		*expected = unknown_expected_context;
 
 	return 0;
 }
@@ -161,23 +168,67 @@ selinux_getfilecon(struct tcb *tcp, const char *path, char **secontext,
 	if (!proc_pid)
 		return -1;
 
-	int rc = -1;
-	char fname[PATH_MAX];
+	/*
+	 * The path may contain .../proc/self/... or .../proc/self<eos>.
+	 * We need to replace 'self' by the target PID, or else we would
+	 * resolve the path with 'self' being strace itself.
+	 * In some corner cases, it's impossible to guess so keep it as is if
+	 * we don't find any occurrence of /proc/self.
+	 * Finally there may be breakage if the pathname contains /proc/self
+	 * but has nothing to do with /proc.
+	 */
 
-	if (path[0] == '/')
+	int rc;
+	char fname[PATH_MAX];
+	char buf[PATH_MAX + 1];
+	size_t l = strlcpy(fname, path, sizeof(fname));
+	if (l >= sizeof(fname) - 1)
+		return -1;
+	int proc_self_substituted = 0;
+
+	for (;;) {
+		char *s = strstr(fname, "/proc/self/");
+		if (s == NULL) {
+			/* Corner case: check if the path ends with /proc/self */
+			s = fname + strlen(fname) - sizeof("/proc/self") + 1;
+			if (strcmp("/proc/self", s))
+				s = NULL;
+		}
+		if (s == NULL)
+			break;
+		*s = '\0';
+		rc = snprintf(buf, sizeof(buf), "%s/proc/%u%s",
+			fname, proc_pid, s + sizeof("/proc/self") - 1);
+		if ((unsigned int) rc >= sizeof(buf))
+			return -1;
+		strcpy(fname, buf);
+		proc_self_substituted = 1;
+	}
+
+	strcpy(buf, fname);
+
+retry:
+	if (buf[0] == '/')
 		rc = snprintf(fname, sizeof(fname), "/proc/%u/root%s",
-			       proc_pid, path);
+			       proc_pid, buf);
 	else if (tcp->last_dirfd == AT_FDCWD)
 		rc = snprintf(fname, sizeof(fname), "/proc/%u/cwd/%s",
-			       proc_pid, path);
+			       proc_pid, buf);
 	else if (tcp->last_dirfd >= 0 )
 		rc = snprintf(fname, sizeof(fname), "/proc/%u/fd/%u/%s",
-			       proc_pid, tcp->last_dirfd, path);
+			       proc_pid, tcp->last_dirfd, buf);
+	else
+		return -1;
 
 	if ((unsigned int) rc >= sizeof(fname))
 		return -1;
 
-	rc = getfilecon(fname, secontext);
+	rc = lgetfilecon(fname, secontext);
+	if ((rc < 0) && (errno == ENOENT) && proc_self_substituted) {
+		proc_self_substituted = 0;
+		strcpy(buf, path);
+		goto retry;
+	}
 	if (rc < 0 || !is_number_in_set(SECONTEXT_MISMATCH, secontext_set))
 		return rc;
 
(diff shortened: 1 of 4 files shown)


src/secontext.c
```diff
@@ -161,44 +168,110 @@ selinux_getfilecon(struct tcb *tcp, const char *path, char **secontext,
...
+	 * In some corner cases, it's impossible to guess so keep it as is if
+	 * we don't find any occurrence of /proc/self.
+	 * Finally there may be breakage if the pathname contains /proc/self
+	 * but has nothing to do with /proc.
+	 */
+
+	int rc;
 	char fname[PATH_MAX];
+	char buf[PATH_MAX + 1];
+	size_t l = strlcpy(fname, path, sizeof(fname));
```
ldv-alt, line 184: strlcpy() was introduced in glibc-2.38, I don't think it's portable enough, and CI confirms that.

src/secontext.c
```diff
@@ -21,6 +22,8 @@
 #include "xmalloc.h"
 #include "xstring.h"
 
+static char unknown_expected_context[] = "??";
```
ldv-alt, line 25: This should be `const`.

src/secontext.c
```diff
@@ -116,18 +119,21 @@ selinux_getfdcon(pid_t pid, int fd, char **secontext, char **expected)
...
 
-	char linkpath[sizeof("/proc/%u/fd/%u") + 2 * sizeof(int)*3];
-	xsprintf(linkpath, "/proc/%u/fd/%u", proc_pid, fd);
+	int rc;
+	char buf[PATH_MAX + 1];
 
-	int rc = getfilecon(linkpath, secontext);
+	rc = snprintf(buf, sizeof(buf), "/proc/%u/fd/%u", proc_pid, fd);
+	if ((unsigned int) rc >= sizeof(buf))
+		return -1;
```
ldv-alt, line 127: I think the use of `linkpath` was justified: if that buffer overflows, it's a programming error, not a runtime error.

src/secontext.c
```diff
@@ -9,6 +9,7 @@
 
 #include <stdlib.h>
 #include <fcntl.h>
+#include <libgen.h>
```
ldv-alt, line 12: I don't think using weird routines declared in this header is justified.

src/secontext.c
```diff
@@ -161,44 +168,110 @@ selinux_getfilecon(struct tcb *tcp, const char *path, char **secontext,
...
+	int rc;
 	char fname[PATH_MAX];
+	char buf[PATH_MAX + 1];
+	size_t l = strlcpy(fname, path, sizeof(fname));
+	if (l >= sizeof(fname) - 1)
+		return -1;
+	int proc_self_substituted = 0;
+
+	for (;;) {
+		char *s = strstr(fname, "/proc/self/");
```
ldv-alt, line 190: Sorry, I don't understand why do we need to substitute all occurrences of /proc/self/ in the string.

ldv-alt (changes requested):

src/secontext.c
rmetrich: Because you could have a query on /proc/self/root/proc/self/root for example. You cannot control how many /proc/self will be found, hence the loop.

src/secontext.c
rmetrich: That's for `dirname()` and `basename()`. I can reimplement these if needed.

src/secontext.c
rmetrich: That was just to save some stack space since there are already many buffers in stack.
Let me revert that, even though there is no chance the size of the /proc link > PATH_MAX.

src/secontext.c
ldv-alt: basename() from <string.h> is better.  I suppose this works:
```
#include <libgen.h>
#undef basename
```

rmetrich: @ldv-alt What do you think of searching for the last slash from the end instead of using `dirname()` and `basename()`.
I was initially fearing that this would require handling escaping, but it seems on Linux at least that it's not possible to have a slash in the file name at all.
It may be on Windows (+ NFS or Samba), but in such case there won't be SELinux labels for sure, so the code will not execute at all.

ldv-alt: I think in this case we can assume that slash is a directory separator.

src/secontext.c
```diff
@@ -137,10 +139,11 @@ selinux_getfdcon(pid_t pid, int fd, char **secontext, char **expected)
 	 * may be reused by a different file with different context.
 	 */
 	strace_stat_t st;
-	if (stat_file(linkpath, &st))
+	if (stat_file(buf, &st) < 0)
 		return 0;
 
-	get_expected_filecontext(buf, expected, st.st_mode);
+	if ((get_expected_filecontext(buf, expected, st.st_mode) < 0) && (errno != ENOENT))
+		*expected = (char *) unknown_expected_context;
```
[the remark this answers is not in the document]
rmetrich: We could, but then `get_execpted_filecontext()` should always return 0, even in case of failure (due to SELinux library for example) and we would lose the reason for failure.

src/secontext.c
```diff
@@ -161,44 +164,123 @@ selinux_getfilecon(struct tcb *tcp, const char *path, char **secontext,
...
+	/*
+	 * We replace all the occurrences of /proc/self[/] because we cannot
+	 * know how many there are (e.g. /proc/self/root/proc/self/root/foo).
+	 */
+	for (;;) {
+		char *s = strstr(fname, "/proc/self/");
+		if (s == NULL) {
+			/* Corner case: check if the path ends with /proc/self */
+			s = fname + strlen(fname) - sizeof("/proc/self") + 1;
+			if (strcmp("/proc/self", s))
```
ldv-alt, line 196: When the string in `fname` isn't long enough, this would result to an out of bounds read.

src/secontext.c
```diff
@@ -161,44 +164,123 @@ selinux_getfilecon(struct tcb *tcp, const char *path, char **secontext,
...
-	if (stat_file(resolved, &st) < 0)
-		goto out;
+	if (S_ISLNK(st.st_mode)) {
+		strcpy(buf, fname);
+		/* Split buf into dirname and basename */
+		char *dirname = buf;
+		char *b;
+		for (b = buf + strlen(buf) - 1; *b != '/'; b--)
+			assert(b > buf); /* we have an absolute path hence at least one / */
+		assert(b > buf + 6); /* we have a dirname (starting with /proc/) */
```
ldv-alt, line 259: Could we simplify this, e.g.
```suggestion
		char *b = strrchr(buf, '/');
		if (!b)
				return 0;
```

src/secontext.c
```diff
@@ -161,44 +164,123 @@ selinux_getfilecon(struct tcb *tcp, const char *path, char **secontext,
...
+		*b = '\0';
+		char *basename = xstrdup(b + 1);
```
ldv-alt, line 261: Why can't this be `const char *basename = b + 1` instead?

src/secontext.c
```diff
@@ -161,44 +164,123 @@ selinux_getfilecon(struct tcb *tcp, const char *path, char **secontext,
...
+		rc = snprintf(buf, sizeof(buf), "%s/%s", resolved, basename);
+		free(resolved);
+		free(basename);
+		if ((unsigned int) rc >= sizeof(buf))
+			return 0;
+	} else {
+		char *resolved = realpath(fname, NULL);
+		if (!resolved)
+			return 0;
+		strcpy(buf, resolved);
```
ldv-alt, line 278: Even if `realpath` follows the spec and doesn't return a string that wouldn't fit the destination buffer, why do we need this copying in the first place?

src/secontext.c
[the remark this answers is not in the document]
ldv-alt: If changing `get_expected_filecontext()` doesn't sound appealing, what about creating a thin wrapper around it that would implement this `unknown_expected_context` assignment?

src/secontext.c
```diff
@@ -176,49 +176,63 @@ selinux_getfilecon(struct tcb *tcp, const char *path, char **secontext,
...
 		if (s == NULL)
 			break;
-		*s = '\0';
-		rc = snprintf(buf, sizeof(buf), "%s/proc/%u%s",
-			fname, proc_pid, s + sizeof("/proc/self") - 1);
-		if ((unsigned int) rc >= sizeof(buf))
-			return -1;
-		strcpy(fname, buf);
-		proc_self_substituted = 1;
+		switch (s[sizeof("/proc/self")]) {
```
ldv-alt, line 196: This is off-by-one, should be strlen or sizeof-1.

src/secontext.c
```diff
@@ -161,44 +164,130 @@ selinux_getfilecon(struct tcb *tcp, const char *path, char **secontext,
...
+			/* path ends with /proc/self */
+			rc = snprintf(buf, sizeof(buf), "/proc/%u", proc_pid);
+			if ((unsigned int) rc >= sizeof(fname) - (s - fname))
+				return -1;
+			strcpy(s, buf);
+			proc_self_substituted = true;
+			goto done_proc_self;
+		case '/':
+			rc = snprintf(buf, sizeof(buf), "/proc/%u/%s",
+				proc_pid, s + sizeof("/proc/self/"));
```
ldv-alt, line 207: `sizeof("/proc/self/")` is off-by-one, should be `strlen("/proc/self/")` or `sizeof("/proc/self/") - 1` or `sizeof("/proc/self")`.

src/secontext.c
```diff
@@ -161,44 +164,130 @@ selinux_getfilecon(struct tcb *tcp, const char *path, char **secontext,
...
+			if ((unsigned int) rc >= sizeof(fname) - (s - fname))
+				return -1;
+			strcpy(s, buf);
+			proc_self_substituted = true;
+			fname_ptr = s + sizeof("/proc");
+			break;
+		default:
+			fname_ptr = s + sizeof("/proc/self");
```
ldv-alt, line 215: off-by-one here as well.

src/secontext.c
```diff
@@ -137,10 +139,11 @@ selinux_getfdcon(pid_t pid, int fd, char **secontext, char **expected)
 	 * may be reused by a different file with different context.
 	 */
 	strace_stat_t st;
-	if (stat_file(linkpath, &st))
+	if (stat_file(buf, &st) < 0)
```
ldv-alt, line 142: Sorry, I don't follow this particular change.  Isn't stat(linkpath) more reliable in this case than stat(buf)?

rmetrich: We want here the context of the target pointed by the symlink `/proc/<pid>/fd/<fd>`, not the symlink itself, hence the stat on `buf`.

ldv-alt: As stat_file() follows symlinks, there shouldn't bean issue, unless the file is deleted, in that case stat(buf) wouldn't be able to access it.

tests/openat.c
```diff
@@ -161,6 +167,158 @@ main(void)
...
+		       sprintrc(fd), target_secontext);
+		if (fd != -1)
+			close(fd);
+	}
+
+	const char *proc_self_paths[] = {
+		"/proc/self", "/proc/self/root/proc/self"
+	};
+
+	for (unsigned int i = 0; i < sizeof(proc_self_paths)/sizeof(*proc_self_paths); i++) {
```
ldv-alt, line 193:
```c
for (unsigned int i = 0; i < ARRAY_SIZE(proc_self_paths); i++) {
```

tests/openat.c
```diff
@@ -161,6 +167,158 @@ main(void)
...
+		 *
+		 * Additionally the SECONTEXT_PID_MY() appends a space,
+		 * instead of prepending it, so fix this here
+		 */
+		char *secontext = SECONTEXT_PID_MY();
+		if (strlen(secontext) > 0) {
+			char *c;
+			for (c = secontext + strlen(secontext) - 1; c > secontext; c--)
+				*c = *(c - 1);
+			*c = ' ';
```
ldv-alt, line 208:
```c
memmove(secontext + 1, secontext, strlen(secontext) - 1);
secontext[0] = ' ';
```

tests/openat.c
```diff
@@ -161,6 +167,158 @@ main(void)
...
+		}
+
+		fd = syscall(__NR_openat, cwd_fd, proc_self_paths[i], O_RDONLY);
+		printf("%s%s(%d%s, \"%s\"%s, O_RDONLY) = %s%s\n",
+		       my_secontext, "openat",
+		       cwd_fd, cwd_secontext,
+		       proc_self_paths[i], secontext,
+		       sprintrc(fd), secontext);
```
ldv-alt, line 216: For some reason this doesn't pass CI on ubuntu-22.04 runners with the following diagnostics:
```diff
-[unconfined] openat(4, "/proc/self" [unconfined], O_RDONLY) = 5 [unconfined]
-[unconfined] openat(4, "/proc/self/root/proc/self" [unconfined], O_RDONLY) = 5 [unconfined]
+[unconfined] openat(4, "/proc/self", O_RDONLY) = 5
+[unconfined] openat(4, "/proc/self/root/proc/self", O_RDONLY) = 5
```

src/secontext.c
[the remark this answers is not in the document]
ldv-alt: By the way, `get_expected_filecontext` currently doesn't set `errno` in case of cached `disabled == true`, leading to undefined behavior.

PR #340, opened 2025-06-21
v4l2: decode `VIDIOC_EXPBUF`

pobrn:
```
Decode the `VIDIOC_EXPBUF` v4l2 ioctl as follows:

  ioctl(17</dev/video10>,
        VIDIOC_EXPBUF,
        {type=V4L2_BUF_TYPE_VIDEO_CAPTURE, index=2, plane=0, flags=O_RDWR|O_CLOEXEC}
          => {fd=25</dmabuf:>}) = 0 <0.000091>

Every current field is supported, the `fd` field is decoded on return.

* src/v4l2.c: Decode `VIDIOC_EXPBUF`.
* tests/ioctl_v4l2.c: Remove `VIDIOC_EXPBUF` from list of unsupported ioctls, and add some `VIDIOC_EXPBUF` tests.
```

PS: I couldn't subscribe to the mailing list, so here it is.

### DIFF
--- a/tests/ioctl_v4l2.c
+++ b/tests/ioctl_v4l2.c
@@ -12,6 +12,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <sys/ioctl.h>
+#include "kernel_fcntl.h"
 #include "kernel_v4l2_types.h"
 
 #define cc0(arg) ((unsigned int) (unsigned char) (arg))
@@ -499,7 +500,6 @@ main(void)
 		const char *str;
 	} unsupp_cmds[] = {
 		{ ARG_STR(VIDIOC_OVERLAY) },
-		{ ARG_STR(VIDIOC_EXPBUF) },
 		{ ARG_STR(VIDIOC_G_AUDIO) },
 		{ ARG_STR(VIDIOC_S_AUDIO) },
 		{ ARG_STR(VIDIOC_QUERYMENU) },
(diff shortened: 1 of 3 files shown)


tests/ioctl_v4l2.c
```diff
@@ -1334,6 +1333,27 @@ main(void)
...
+	struct v4l2_exportbuffer *const p_v4l2_expbuf =
+	       page_end - sizeof(*p_v4l2_expbuf);
+	ioctl(-1, VIDIOC_EXPBUF, p_v4l2_expbuf);
+	printf("ioctl(-1, %s, {"
+	       "type=%#x" NRAW(" /* V4L2_BUF_TYPE_??? */") ", "
+	       "index=%u, "
+	       "plane=%u, "
+	       "flags="
+	       NABBR("0xcfcecdcc") VERB(" /* ")
+	       NRAW("O_RDONLY|O_CREAT|O_EXCL|O_NOCTTY|O_APPEND|O_NONBLOCK|O_DIRECT|O_LARGEFILE|O_NOFOLLOW|O_NOATIME|O_CLOEXEC|__O_TMPFILE|0xcf80000c")
```
ldv-alt, line 1349: I'm afraid hardcoding this is not portable as values for some of these constants, e.g. O_LARGEFILE, differ between architectures.

ldv-alt: The test fails with
```diff
-ioctl(-1, VIDIOC_EXPBUF, {type=0xffffffff /* V4L2_BUF_TYPE_??? */, index=123, plane=321, flags=O_CLOEXEC}) = -1 EBADF (Bad file descriptor)
+ioctl(-1, VIDIOC_EXPBUF, {type=0xffffffff /* V4L2_BUF_TYPE_??? */, index=123, plane=321, flags=O_RDONLY|O_CLOEXEC}) = -1 EBADF (Bad file descriptor)
```
Looks like it might be a good idea to do `p_v4l2_expbuf->flags = O_RDONLY|O_CLOEXEC` and `NRAW("O_RDONLY|O_CLOEXEC")`.

ldv-alt (approved):

pobrn:
Thank you! And sorry, I must have missed the other file with tests.

PR #347, opened 2025-08-12
tests: make uio test regex work with musl

### LINKED ISSUE #346: uio test fails with musl-libc
When running the test suite on a system with musl-libc, the uio test fails:

```
Failed patterns of expected output:
#3: pwrite(64)?\(0, "\\0\\0\\0\\0", 4, 1004211379570065135\) += 4
#4: pwritev\(0, \[\{iov_base="\\0\\0\\0\\0", iov_len=4\}\], 1, 1004211379570065135\) += 4

Actual output:
mmap(0x56381027a000, 4096, PROT_NONE, MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS, -1, 0) = 0x56381027a000
close(0)                                = 0
open("/dev/zero", O_RDONLY|O_LARGEFILE) = 0
pread64(0, "\0\0\0\0", 4, 1004211379570065135) = 4
preadv(0, [{iov_base="\0\0\0\0", iov_len=4}], 1, 1004211379570065135) = 4
close(0)                                = 0
open("/dev/null", O_WRONLY|O_LARGEFILE) = 0
pwritev2(0, [{iov_base="\0\0\0\0", iov_len=4}], 1, 1004211379570065135, RWF_NOAPPEND) = 4
pwritev2(0, [{iov_base="\0\0\0\0", iov_len=4}], 1, 1004211379570065135, RWF_NOAPPEND) = 4
close(0)                                = 0
+++ exited with 0 +++
uio.test: failed test: ../../src/strace -e%desc ../uio output mismatch
FAIL uio.test (exit status: 1)
```

The reason seems to be the way musl handles the pwrite calls: both `pwrite` and `pwritev` are actually `pwritev2`:
calling `pwrite` calls `pwritev2`: https://git.musl-libc.org/cgit/musl/tree/src/unistd/pwrite.c#n10
calling `pwritev` calls `pwritev2`: https://git.musl-libc.org/cgit/musl/tree/src/unistd/pwritev.c#n10

OldManYellsAtCloud:
musl-libc handles the `pwrite` calls differently than glibc, making the uio tests fail. Using glibc, there are separated `pwrite` and `pwritev` calls, however musl uses only `pwritev2` for both `pwrite` and `pwritev`.

To account to this difference, change the verifying regex in way to match both glibc and musl outputs.

Closes #346


ldv-alt:
While the issue looks like musl-specific, it can happen with any libc when it switches to a different set of system calls.
It seems to me that the right fix for this particular test would be to update the regexp to take pwritev2 into account.

OldManYellsAtCloud:
Now that I did a test with your suggestion, I noticed that there are other `pwrite` tests failing similarly too with musl - will check them too while I'm already at it.

ldv-alt (approved):

kraj:
@OldManYellsAtCloud thanks for fixing it, There are following failures on musl which are due to exact same problem - 

```
Failed ptests: {'strace': ['preadv-pwritev.gen.test', 'pread64-pwrite64.gen.test', 'pwritev.gen.test', 'uio.test', 'pread64-pwrite64.gen', 'preadv-pwritev.gen', 'pwritev.gen', 'uio']}
```

OldManYellsAtCloud:
@kraj Yes - those tests are a bit different, and unfortunately I found myself fighting random fires in other projects, and had to re-prioritize some things. It's on my list to fix all tests w/ musl in strace... but if you have time, feel free to check also.

kraj:
> @kraj Yes - those tests are a bit different, and unfortunately I found myself fighting random fires in other projects, and had to re-prioritize some things. It's on my list to fix all tests w/ musl in strace... but if you have time, feel free to check also.

Perhaps the testcase could call syscalls directly instead of depending upon libc wrapper functions can make them portable.

PR #345, opened 2025-08-12
kvm: dump kvm_run area

masatake:
strae has `-e kvm=vcpu` already.
This pull request extends the option.
This pull request adds `-e kvm=vcpu+` option.
The new option is for dumping the kvm_run mmap'ed area.
```console
[yamato@dev64]~/var/strace/tests% ../src/strace -y -e ioctl -e kvm=vcpu+ ./ioctl_kvm_run_auxstr_vcpu
...
ioctl(5<anon_inode:kvm-vcpu:0>, KVM_RUN, 0) = 0 (KVM_EXIT_IO)
VCPU>{request_interrupt_window=0, immediate_exit=0, 
VCPU+ exit_reason=2 (KVM_EXIT_IO), ready_for_interrupt_injection=0, if_flag=0, flags=0, 
VCPU+ cr8=0000000000000000, apic_base=0000000000000000 => 0x000000fee00900, 
VCPU+ {io={direction=KVM_EXIT_IO_OUT, port=0x3f8, count=1, data_offset=0x00000000001000}}}
ioctl(5<anon_inode:kvm-vcpu:0>, KVM_RUN, 0) = 0 (KVM_EXIT_HLT)
VCPU>{request_interrupt_window=0, immediate_exit=0, 
VCPU+ exit_reason=5 (KVM_EXIT_HLT), ready_for_interrupt_injection=0, if_flag=0, flags=0, 
VCPU+ cr8=0000000000000000, apic_base=0x000000fee00900}
+++ exited with 0 +++
```

VCPU> and VCPU+ are the prefixes for the dump data.

PR #357, opened 2025-10-29
kvm:  decode before and after states

masatake:
@ldv-alt I'm very sorry to make a pull request just after you merged #345.

However, I found a limitation in #345 that we cannot ignore.

I have tried the merged version of strace with the tracee available at https://github.com/masatake/kernel-essential-training/blob/master/kvm/add-iomem-kvm.c.
add-iomem-kvm.c demonstrates device emulation via I/O memory.

The tracee reads data from the address that has no memory-mapping.
In such a case, the tracee gains control with KVM_EXIT_IOMEM.
The tracee provides data via `run->mmio.data[0] = ` here `run` points memory region where kvm_run is mapped. Then the tracee invokes ioctl(KVM_RUN) to enter the guest mode.

I proposed `VCPU> ` in #345; strace mainly decoded the after-state of the kvm_run structure.
Though I considered decoding the difference between the before-state and after-state using tprint_value_changed, I found that my code was insufficient. I assumed exit_reason didn't change between the before-state and the after-state. e.g., When the guest exits with KVM_EXIT_IO, strace cannot decode the value of run->mmio.data[0] because the handler for KVM_EXIT_IOMEM decodes run->mmio.data[0].

In this pull request, I propose `VCPU< ` for decoding before-state and `VCPU> ` for decoding after-state.

```console
ioctl(5, KVM_RUN, 0)                    = 0 (KVM_EXIT_IO)
 VCPU:0< {..., exit_reason=6 /* KVM_EXIT_MMIO */, ..., {mmio={..., data=[0x3, 0, 0, 0, 0, 0, 0, 0], len=1, is_write=0}}}
 VCPU:0> {..., exit_reason=2 /* KVM_EXIT_IO */, ..., {io={...}}}
```

Thank you for reading.

### DIFF
--- a/tests/ioctl_kvm_run_common.c
+++ b/tests/ioctl_kvm_run_common.c
@@ -82,6 +82,10 @@ __asm__(
 	"	out %al, (%dx)		\n"
 	"	mov $0x2000, %dx	\n"
 	"	movb $0xdf, (%edx)	\n"
+	"	mov $0x2001, %dx	\n"
+	"	movb (%edx), %al	\n" /* read from iomem */
+	"	mov $0x03f9, %dx	\n"
+	"	out %al, (%dx)		\n" /* use ioport again */
 	"	hlt			\n"
 # ifdef __x86_64__
 	".code64			\n"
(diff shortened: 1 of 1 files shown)


tests/ioctl_kvm_run_auxstr_vcpu.c
```diff
@@ -14,28 +14,34 @@
...
-		default: str = "???";
+		default: return "???";
 	}
+}
 
+static void
+print_KVM_RUN(const int fd, const char *const dev,
+	      const struct kvm_run *run_before, const struct kvm_run *run_after)
+{
+	const char *str = decode_exit_reason(run_after->exit_reason);
```
ldv-alt: `str_after` would be more consistent.

masatake:
Updated.

ldv-alt (approved):

masatake:
Thank you.

PR #364, opened 2025-11-08
kvm: decode data written to and read from I/O ports

masatake:
With the "-e kvm=vcpu+" option, strace now decodes data written to and read from I/O ports.

* src/kvm.c (kvm_run_structure_decode_io): Decode data written to or read from I/O ports as a byte array.
(kvm_run_structure_decode_main): Add struct vcpu_info as a new parameter and pass it down to kvm_run_structure_decode_io(). (kvm_run_structure_decode): Pass a struct vcpu_info object to kvm_run_structure_decode_main.
* tests/ioctl_kvm_run_common.c (run_kvm): Allocate buffer for run_before dynamically and copy entire mmap'ed region to the buffer.
* tests/ioctl_kvm_run_auxstr_vcpu_more.c (print_kvm_run_more): Print data written to or read from I/O ports.

### DIFF
--- a/tests/ioctl_kvm_run_auxstr_vcpu_more.c
+++ b/tests/ioctl_kvm_run_auxstr_vcpu_more.c
@@ -36,11 +36,16 @@ print_kvm_run_more(const char *prefix, const char *reason_str, const struct kvm_
 	switch (run->exit_reason) {
 	case KVM_EXIT_IO:
 		printf(", {io={direction=%s"
-		       ", size=%u, port=%#04x, count=%u, data_offset=%#016llx}}",
+		       ", size=%u, port=%#04x, count=%u, data_offset=%#016llx",
 		       run->io.direction == KVM_EXIT_IO_IN
 		       ? "KVM_EXIT_IO_IN" : "KVM_EXIT_IO_OUT",
 		       run->io.size, run->io.port,
 		       run->io.count, run->io.data_offset);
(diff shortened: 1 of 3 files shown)


tests/ioctl_kvm_run_auxstr_vcpu_more.c
```diff
@@ -36,11 +36,16 @@ print_kvm_run_more(const char *prefix, const char *reason_str, const struct kvm_
...
-		       ", size=%u, port=%#04x, count=%u, data_offset=%#016llx}}",
+		       ", size=%u, port=%#04x, count=%u, data_offset=%#016llx",
 		       run->io.direction == KVM_EXIT_IO_IN
 		       ? "KVM_EXIT_IO_IN" : "KVM_EXIT_IO_OUT",
 		       run->io.size, run->io.port,
 		       run->io.count, run->io.data_offset);
+		fputs(", data=[", stdout);
+		for (unsigned long i = 0; i < (run->io.size * run->io.count); ++i)
+			printf("%s0x%x", i == 0?  "" : ", ",
+			       ((unsigned char *)run + run->io.data_offset)[i]);
```
ldv-alt: Since print_xint_array_member() uses PRINT_VAL_X(), one should use %#x instead of 0x%x to avoid differences in printing of 0.

ldv-alt (approved):

Issue #344, opened 2025-08-05
propagate unix socket name from bind()


k4lizen:
This is for the `--decode-fds=all` option. Currently after a unix socket gets bind()ed a name:
```
bind(103<UNIX:[1629885]>, {sa_family=0x1 /* AF_UNIX */, sun_path=@"my-neat-socket\x00\x00\x00"}, 110 <unfinished ...>
```
that name is not shown in other syscalls involving that fd:
```
listen(103<UNIX:[1629885]>, 4 <unfinished ...>
```
This would be useful because when analyzing a binary you may want to search for all syscalls involving operations on a given socket, and the name of the socket stays consistent between program runs (or if the socket is destroyed/recreated).

ldv-alt:
It's shown as soon as the kernel shows it.

k4lizen:
i don't understand what you mean.
1. how does the kernel show it?
2. you know the name of the socket from the syscall argument. you can just save it and reuse it. no need for kernel API, no?

ldv-alt:
> i don't understand what you mean.
> 
>     1. how does the kernel show it?

Via netlink SOCK_DIAG_BY_FAMILY interface.

>     2. you know the name of the socket from the syscall argument. you can just save it and reuse it. no need for kernel API, no?

strace doesn't imagine, it just shows what actually happens between the kernel and userspace, and after successful `bind` syscall the kernel doesn't report the socket as connected to a specific address.

In strace, we have [a test](https://github.com/strace/strace/blob/master/tests/net-yy-unix.c#L67) that covers this case.

masatake:
I was unable to reproduce the issue on Fedora 42.
```
[yamato@dev64]~% strace -e bind,listen --decode-fds=all nc -U -l /tmp/my.sock
bind(3<UNIX-STREAM:[863696]>, {sa_family=AF_UNIX, sun_path="/tmp/my.sock"}, 14) = 0
listen(3<UNIX-STREAM:[863696,"/tmp/my.sock"]>, 10) = 0
```

I was able to reproduce the issue on Raspberry Pi 5, where Debian GNU/Linux 13 (trixie) run.
```
$ strace -e bind,listen -yy nc -U -l /tmp/my.sock
bind(3<UNIX-STREAM:[288445]>, {sa_family=AF_UNIX, sun_path="/tmp/my.sock"}, 15) = 0
listen(3<UNIX-STREAM:[288445]>, 5)      = 0
```

[...]

PR #379, opened 2026-02-27
[RFC] utilize pidfd_getfd in decoding sockets

masatake:
See #344 the idea behind this pull request.


masatake:
I should prepare a real pull request implementing something interesting.

PR #382, opened 2026-03-10
add rseq syscall decoding support

jirislaby:


ldv-alt (changes requested):
All tests that use ubuntu-22.04 image without fresh kernel headers failed to compile because of missing <linux/rseq.h>, looks like this header has to be bundled.

jirislaby:
> All tests that use ubuntu-22.04 image without fresh kernel headers failed to compile because of missing <linux/rseq.h>, looks like this header has to be bundled.

Could you help fixing this ^^^, as I have no clue how to bundle the header.

ldv-alt (approved):

PR #385, opened 2026-03-19
udmabuf: decode ioctls

pobrn:
Decode the two currently supported udmabuf ioctls: `UDMABUF_CREATE`, and `UDMABUF_CREATE_LIST`. Every field is supported.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,8 @@ Noteworthy changes in release ?.?? (????-??-??)
 ===============================================
 
 * Improvements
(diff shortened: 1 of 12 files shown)


pobrn:
Should `udmabuf.h` be bundled? It has remained unchanged since it was first added in linux 4.20. What is the policy?

ldv-alt (changes requested):

ldv-alt (approved):

PR #387, opened 2026-04-03
v4l2: decode `VIDIOC_QUERYMENU`

pobrn:
Decode the `VIDIOC_QUERYMENU` ioctl, which is used for both string and integer menu controls. Unfortunately without querying the type of the control it is not possible to know if it is a string or integer, so print both.

Integer menus:

  ioctl(3, VIDIOC_QUERYMENU, {id=V4L2_CTRL_CLASS_USER+0xf907, index=2} => {name="\2", value=2}) = 0 <0.000025>

String menus:

  ioctl(3, VIDIOC_QUERYMENU, {id=V4L2_CTRL_CLASS_USER+0xf904, index=1} => {name="Menu Item 1", value=7310548498887107917}) = 0 <0.000039>

* src/v4l2.c: Decode `VIDIOC_QUERYMENU`.
* tests/ioctl_v4l2.c: Remove `VIDIOC_QUERYMENU` from the list of unsupported ioctls, and add some `VIDIOC_EXPBUF` tests.

### DIFF
--- a/src/v4l2.c
+++ b/src/v4l2.c
@@ -1147,6 +1147,36 @@ print_v4l2_queryctrl(struct tcb *const tcp, const kernel_ulong_t arg)
 }
 
 static int
+print_v4l2_querymenu(struct tcb *const tcp, const kernel_ulong_t arg)
+{
+	struct v4l2_querymenu qm;
+
+	if (entering(tcp)) {
+		tprints_arg_next_name("argp");
+		if (umove_or_printaddr(tcp, arg, &qm))
+			return RVAL_IOCTL_DECODED;
+
+		tprint_struct_begin();
+		PRINT_FIELD_V4L2_CID(qm, id, false);
+		tprint_struct_next();
+		PRINT_FIELD_U(qm, index);
+
+		return 0;
+	}
+
+	if (!syserror(tcp) && !umove(tcp, arg, &qm)) {
+		tprint_struct_end_value_changed_struct_begin();
+		PRINT_FIELD_CSTRING(qm, name);
+		tprint_struct_next();
+		PRINT_FIELD_D(qm, value);
+	}
+
+	tprint_struct_end();
+
+	return RVAL_IOCTL_DECODED;
+}
+
+static int
 print_v4l2_query_ext_ctrl(struct tcb *const tcp, const kernel_ulong_t arg)
 {
 	struct v4l2_query_ext_ctrl c;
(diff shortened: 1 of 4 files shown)


pobrn:
I get an error 400 when trying to check why the code coverage test fails: https://app.codecov.io/gh/strace/strace/pull/387/indirect-changes

src/v4l2.c
```diff
@@ -1146,6 +1146,36 @@ print_v4l2_queryctrl(struct tcb *const tcp, const kernel_ulong_t arg)
...
+		PRINT_FIELD_V4L2_CID(qm, id, false);
+		tprint_struct_next();
+		PRINT_FIELD_U(qm, index);
+
+		return 0;
+	}
+
+	if (!syserror(tcp) && !umove(tcp, arg, &qm)) {
+		tprint_struct_end_value_changed_struct_begin();
+		PRINT_FIELD_STRING(qm, name, sizeof(qm.name), QUOTE_0_TERMINATED);
```
ldv-alt: Shouldn't `PRINT_FIELD_CSTRING` be used here instead of `PRINT_FIELD_STRING`?

tests/ioctl_v4l2-success.c
```diff
@@ -1735,6 +1735,36 @@ main(int argc, char **argv)
...
+	qmenu->id = V4L2_CID_EXPOSURE_AUTO;
+	strcpy((char *) qmenu->name, "Aperture Priority Mode");
+
+	ioctl(-1, VIDIOC_QUERYMENU, qmenu);
+	printf("ioctl(-1, %s, {"
+	       "id=" NABBR("%#x") VERB(" /* ") NRAW("V4L2_CID_EXPOSURE_AUTO") VERB(" */") ", "
+	       "index=%u"
+	       "} => {"
+	       "name=\"Aperture Priority Mode\", "
+	       "value=7310034288171642945"
```
ldv-alt: Is this value the same on big-endian and little-endian architectures?

ldv-alt (changes requested):

src/v4l2.c
pobrn: I haven't found that for some reason... it's probably better.

tests/ioctl_v4l2-success.c
pobrn: Ahh, of course not. I have updated the test.

ldv-alt (approved):

PR #404, opened 2026-06-07
Give a hint about yama if strace -p fails on EPERM

stanislav-brabec:
If strace -p fails on EPERM and the PID matches, the reason is not clean. Give a hint that yama could block it.

It displays the extended message if the error is EPERM and UID is not 0. There are some "false matches" as advanced checks like UID match would require parsing of /proc/$PID/status.

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -2,6 +2,7 @@
 /bpf_attr_check.c
 /config.h
 /config.h.in
+/disable_ptrace
 /disable_ptrace_get_syscall_info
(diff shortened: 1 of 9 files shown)


stanislav-brabec:
https://docs.kernel.org/admin-guide/LSM/Yama.html

Note about the text: The hint deliberately does not suggest to turn off Yama, as it has a security implications.

If you want to refer to the documentation, feel free to change it.

ldv-alt (approved):

PR #408, opened 2026-06-15
Append id to "syscall suspended" lines

### LINKED ISSUE #152: Feature request: Display the full syscall line on resuming
Currently when tracing multiple processes you could see something like:
```
[pid  1905] select(0, NULL, NULL, NULL, {0, 198904} <unfinished ...>
[pid 23966] ftruncate(28, 25165824)     = 0
[pid 32266] epoll_wait(20,  <unfinished ...>
[pid  1905] <... select resumed> )      = 0 (Timeout)
```
So on resuming you often need to scroll back to understand of what syscall was continued. especially when you see that resumed syscall failed with some error. So may be there could be an option, so the output could be something like this:
```
[pid  1905] select(0, NULL, NULL, NULL, {0, 198904} <unfinished ...>
[pid 23966] ftruncate(28, 25165824)     = 0
[pid 32266] epoll_wait(20,  <unfinished ...>
[pid  1905] <resumed> select(0, NULL, NULL, NULL, {0, 198904})      = 0 (Timeout)
```
It would make reading the logs easier.

valdaarhun:
While this doesn't print the syscall line in its entirety on resumption, it makes it easier to match suspension-resumption lines when perusing strace's output.

Resolves #152 .

### DIFF
--- a/src/strace.c
+++ b/src/strace.c
@@ -175,6 +175,7 @@ static bool open_append;
 
 struct tcb *printing_tcp;
 static struct tcb *current_tcp;
+static unsigned long resume_id;
 
 struct tcb_wait_data {
 	enum trace_event te; /**< Event passed to dispatch_event() */
@@ -843,6 +844,15 @@ line_ended(void)
 	}
 }
 
+static unsigned long
+update_resume_id(struct tcb *tcp)
+{
+	if (!tcp->resume_id)
+		tcp->resume_id = ++resume_id;
+
+	return tcp->resume_id;
+}
+
 static void
 set_current_tcp(const struct tcb *tcp)
 {
@@ -886,7 +896,7 @@ printleader(struct tcb *tcp)
 			 */
 			set_current_tcp(printing_tcp);
 			tprint_space();
-			tprints_string("<unfinished ...>");
+			tprintf_string("<unfinished #%lu ...>", update_resume_id(printing_tcp));
 			tprint_newline();
 			printing_tcp->curcol = 0;
 		}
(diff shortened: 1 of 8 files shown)


src/syscall.c
```diff
@@ -755,7 +758,9 @@ print_syscall_resume(struct tcb *tcp)
 	    || (tcp->flags & TCB_REPRINT)) {
 		tcp->flags &= ~TCB_REPRINT;
 		printleader(tcp);
-		tprintf_string("<... %s resumed>", tcp_sysent(tcp)->sys_name);
+		tprintf_string("<... %s resumed #%lu>", tcp_sysent(tcp)->sys_name, tcp->resume_id);
```
ldv-alt, line 761: `resume_id` may be 0 when printed for `TCB_REPRINT` case without prior unfinished mark.
In threaded execve when `execve_thread->curcol==0`: `maybe_switch_tcbs` sets `TCB_REPRINT` without calling `update_resume_id`, then `print_syscall_resume` prints `'resumed #0'` instead of a valid ID, creating output that doesn't match any `'<unfinished #0 ...>'` message.

src/strace.c
```diff
@@ -1074,6 +1084,7 @@ alloctcb(int pid)
 			memset(tcp, 0, sizeof(*tcp));
 			list_init(&tcp->wait_list);
 			tcp->pid = pid;
+			tcp->resume_id = 0;
```
ldv-alt, line 1087: Redundant zero initialization of `tcp->resume_id` after memset.

src/strace.c
```diff
@@ -3682,7 +3608,7 @@ print_event_exit(struct tcb *tcp)
 	    && printing_tcp->curcol != 0 && !printing_tcp->staged_output_data) {
 		set_current_tcp(printing_tcp);
 		tprint_space();
-		tprints_string("<unfinished ...>");
+		tprintf_string("<unfinished #%lu ...>", update_resume_id(printing_tcp));
```
ldv-alt, line 3611: Lines 899 and 3611 contain identical code:
`tprintf_string("<unfinished #%lu ...>", update_resume_id(printing_tcp))`.
If the format or logic changes, both locations need updating, risking inconsistency.

ldv-alt (changes requested):

PR #413, opened 2026-07-11
setns: decode namespace transition with --namespace=new option even if nstype is 0

masatake:
When setns is called with nstype specified as 0, any type of namespace
can be joined and the type is determined by the passed fd. Query the
namespace type of the fd by reading the /proc/$pid/fd/$fd symlink to
retrieve the type of namespace being joined.

* NEWS: Mention this change.
* src/clone.c (read_namespace_type_for_fd): New helper function.
(get_namespace_auxstr): Add expected_ns_str parameter.
If flags is 0, print the namespace ID of expected_ns_str.
(SYS_FUNC(clone), SYS_FUNC(clone3), SYS_FUNC(unshare)): Pass NULL
as expected_ns_str to get_namespace_auxstr.
(SYS_FUNC(setns)): Query the namespace type of the joined fd
using read_namespace_type_for_fd if nstype is 0, and pass it as
expected_ns_str to get_namespace_auxstr.
* tests/setns-report-ns-id.c (main): Add a test case for setns(fd, 0).

Assisted-by: Antigravity:Gemini-3.5-Flash (comit log)

### DIFF
--- a/src/clone.c
+++ b/src/clone.c
@@ -81,7 +81,8 @@ read_namespace_id(int pid, const char *const ns_type)
 
 static const char *
 get_namespace_auxstr(int pid, uint64_t flags,
-		     struct tcb *const tcp_for_pid_translation)
+		     struct tcb *const tcp_for_pid_translation,
+		     const char *expected_ns_str)
 {
 	static char str[sizeof("cgroup:[4026531835], "
 			          "ipc:[4026531839], "
@@ -115,6 +116,16 @@ get_namespace_auxstr(int pid, uint64_t flags,
 	}
 
 	*p = '\0';
+
+	if (flags == 0 && expected_ns_str) {
+		const char *ns_name = read_namespace_id(pid, expected_ns_str);
+		if (ns_name) {
+			(void)xappendstr(str, p, "%s", ns_name);
+			return str;
+		}
+		return NULL;
+	}
+
 	for (size_t i = 0; i < ARRAY_SIZE(ns); ++i) {
 		if (flags & ns[i].flags) {
 			const char *ns_name = read_namespace_id(pid, ns[i].str);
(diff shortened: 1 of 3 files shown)


src/clone.c
```diff
@@ -393,6 +406,25 @@ SYS_FUNC(clone3)
...
+	char linkpath[sizeof(fd_path) + sizeof(int) * 3 + sizeof(int) * 3];
+	xsprintf(linkpath, fd_path, pid, fd);
+
+	static char buf[PATH_MAX + 1];
+	ssize_t n = readlink(linkpath, buf, sizeof(buf));
+	if ((size_t) n >= sizeof(buf))
+		return NULL;
+
+	/* buf refers a string like "ns:[4026531838]". */
+	char *p = strchr(buf, ':');
```
ldv-alt, line 423: This fails to null-terminate `buf` after `readlink()`, unlike the similar `read_namespace_id` function which correctly does `buf[n] = '\0'`. As result, subsequent `strchr` call will scan past the end of the valid data (of length `n`) into uninitialized or stale memory.

ldv-alt (changes requested):

src/clone.c
masatake: Updated.

masatake:
I temporarily deleted the entry for this change from NEWS.
I will restore the entry after Dmitry's "Post-release administrivia."

src/clone.c
```diff
@@ -393,6 +406,26 @@ SYS_FUNC(clone3)
...
+	char linkpath[sizeof(fd_path) + sizeof(int) * 3 + sizeof(int) * 3];
+	xsprintf(linkpath, fd_path, pid, fd);
+
+	static char buf[PATH_MAX + 1];
+	ssize_t n = readlink(linkpath, buf, sizeof(buf));
+	if ((size_t) n >= sizeof(buf))
+		return NULL;
+	buf[n] = '\0';
+
+	/* buf refers a string like "ns:[4026531838]". */
```
ldv-alt: Grammar issue: `buf` is not a pointer but an array, so `buf contains a string` rather than `buf refers to a string`.

PR #414, opened 2026-07-20
time: handle dynamic and auxiliary clocks

t-8ch:
Linux supports non-standard auxiliary clocks.
These are identified by their clock IDs.

### DIFF
--- a/src/time.c
+++ b/src/time.c
@@ -256,10 +256,10 @@ printclockname(int clockid)
 		if (xlat_verbose(xlat_verbosity) == XLAT_STYLE_VERBOSE)
 			tprint_comment_begin();
 
-		if ((clockid & CLOCKFD_MASK) == CLOCKFD)
+		if ((clockid & CLOCKFD_MASK) == CLOCKFD) {
 			tprints_fn_begin("FD_TO_CLOCKID");
 			PRINT_VAL_D(CLOCKID_TO_FD(clockid));
-		else {
+		} else {
 			tprints_fn_begin(CPUCLOCK_PERTHREAD(clockid) ?
 					  "MAKE_THREAD_CPUCLOCK" :
 					  "MAKE_PROCESS_CPUCLOCK");
--- a/src/time.c
+++ b/src/time.c
@@ -239,13 +239,24 @@ SYS_FUNC(adjtimex64)
 
 #include "xlat/clockflags.h"
 #include "xlat/clocknames.h"
+#include "xlat/cpuclocknames.h"
+
+#ifndef CLOCKID_TO_FD
+# define CPUCLOCK_PID(clock)		((pid_t) ~((clock) >> 3))
+# define CPUCLOCK_PERTHREAD(clock) \
+	(((clock) & (clockid_t) CPUCLOCK_PERTHREAD_MASK) != 0)
+
+# define CLOCKID_TO_FD(clk)	((unsigned int) ~((clk) >> 3))
+# define CPUCLOCK_CLOCK_MASK	3
+# define CPUCLOCK_PERTHREAD_MASK	4
+# define CLOCKFD_MASK		(CPUCLOCK_PERTHREAD_MASK|CPUCLOCK_CLOCK_MASK)
+# define CPUCLOCK_MAX		3
+# define CLOCKFD			CPUCLOCK_MAX
+#endif
 
 static void
 printclockname(int clockid)
 {
-#ifdef CLOCKID_TO_FD
-# include "xlat/cpuclocknames.h"
-
 	if (clockid < 0) {
 		if (xlat_verbose(xlat_verbosity) != XLAT_STYLE_ABBREV)
 			PRINT_VAL_D(clockid);
@@ -272,8 +283,8 @@ printclockname(int clockid)
 
 		if (xlat_verbose(xlat_verbosity) == XLAT_STYLE_VERBOSE)
 			tprint_comment_end();
+
 	} else
-#endif
 		printxval(clocknames, clockid, "CLOCK_???");
 }
 
--- a/src/time.c
+++ b/src/time.c
@@ -8,6 +8,7 @@
  * SPDX-License-Identifier: LGPL-2.1-or-later
  */
 
+#include "auxiliary_clock.h"
 #include "defs.h"
 #include "kernel_fcntl.h"
 #include <signal.h>
@@ -279,6 +280,28 @@ printclockname(int clockid)
 		}
 		tprint_fn_end();
 		tprint_comment_end();
+
+	} else if (is_auxiliary_clock(clockid)) {
+		if (xlat_verbose(xlat_verbosity) != XLAT_STYLE_ABBREV)
+			PRINT_VAL_X(clockid);
+
+		if (xlat_verbose(xlat_verbosity) == XLAT_STYLE_RAW)
+			return;
+
+		if (xlat_verbose(xlat_verbosity) == XLAT_STYLE_VERBOSE)
+			tprint_comment_begin();
+		else
+			STRACE_PRINT_COLOR_SEQ(COLOR_CONST);
+
+		STRACE_PRINTS("CLOCK_AUX");
+
+		if (xlat_verbose(xlat_verbosity) != XLAT_STYLE_VERBOSE)
+			STRACE_PRINT_COLOR_SEQ(COLOR_ARGVAL);
+
+		STRACE_PRINTF(" + %u", auxiliary_clock_num(clockid));
+
+		if (xlat_verbose(xlat_verbosity) == XLAT_STYLE_VERBOSE)
+			tprint_comment_end();
 
 	} else
 		printxval(clocknames, clockid, "CLOCK_???");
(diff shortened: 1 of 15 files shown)


esyr:
Where do these definitions come from? Linux defines only `CLOCK_AUX` and `MAX_AUX_CLOCK`/`CLOCK_AUX_LAST`. If anything, this should be decoded similarly to `CLOCKFD`, in `printclockname`, I guess, with the latter used where appropriate. Also, that would require test updates.

t-8ch:
@esyr Will do.

t-8ch:
@esyr Does the new version look better?

src/time.c
```diff
@@ -274,6 +275,22 @@ printclockname(int clockid)
...
+		if (xlat_verbose(xlat_verbosity) != XLAT_STYLE_ABBREV)
+			PRINT_VAL_D(clockid);
+
+		if (xlat_verbose(xlat_verbosity) == XLAT_STYLE_RAW)
+			return;
+
+		STRACE_PRINT_COLOR_SEQ(COLOR_CONST);
+		STRACE_PRINTS("CLOCK_AUX");
+		STRACE_PRINT_COLOR_SEQ(COLOR_ARGVAL);
+		STRACE_PRINTF(" + %d", clockid - CLOCK_AUX);
```
ldv-alt: Missing XLAT_STYLE_VERBOSE comment wrapper for CLOCK_AUX symbolic output.

In case of `strace -Xverbose`, `clock_gettime(CLOCK_AUX + 0, ...)` prints `'16 CLOCK_AUX + 0'` instead of `'16 /* CLOCK_AUX + 0 */'`, breaking xlat verbosity convention used by negative clockid handler.

tests/clock_xettime-common.c
```diff
@@ -111,6 +112,16 @@ main(void)
...
 
+#if defined(CLOCK_AUX)
+	k_gettime(CLOCK_AUX + 0, NULL);
+	printf("%s(CLOCK_AUX + 0, NULL) = %s\n",
+	       SYSCALL_NAME_gettime, errstr);
+
+	k_gettime(CLOCK_AUX + 7, NULL);
+	printf("%s(CLOCK_AUX + 7, NULL) = %s\n",
+	       SYSCALL_NAME_gettime, errstr);
+#endif
```
ldv-alt: Test covers `CLOCK_AUX + 7` (last valid) but not `CLOCK_AUX_LAST + 1` (first invalid):
an off-by-one error where upper bound is `'<= CLOCK_AUX + 8'` instead of `'<= CLOCK_AUX_LAST'` would pass tests; value 24 would incorrectly print as `'CLOCK_AUX + 8'` instead of `'CLOCK_???'`.

No test coverage for -Xverbose, -Xraw, or -Xabbrev modes with CLOCK_AUX values:
missing VERBOSE comment wrapper goes undetected; with `-Xverbose`, output is `'16 CLOCK_AUX + 0'` instead of `'16 /* CLOCK_AUX + 0 */'`, causing user confusion or test mismatch if xlat modes are tested.

ldv-alt (changes requested):
Also, user-visible change (CLOCK_AUX decoding support) requires NEWS update.

t-8ch:
Updated with the latest review comments.
Locally I get a bunch of (unrelated) test failures. Could you enable the CI on this PR?

t-8ch:
The tests look better now.
I am a bit unhappy about the need to spread the changes to `clock_xettime` around so much.
What do you think about adding a new dedicated `clockid` test which only tests clock ID decoding?

Also the "Post-release administrivia" commit for the 7.2 release seems to be missing.

src/time.c
```diff
@@ -274,6 +275,26 @@ printclockname(int clockid)
...
+		if (xlat_verbose(xlat_verbosity) != XLAT_STYLE_ABBREV)
+			PRINT_VAL_X(clockid);
+
+		if (xlat_verbose(xlat_verbosity) == XLAT_STYLE_RAW)
+			return;
+
+		if (xlat_verbose(xlat_verbosity) == XLAT_STYLE_VERBOSE)
+			tprint_comment_begin();
+
+		STRACE_PRINT_COLOR_SEQ(COLOR_CONST);
```
esyr: That would blow up inside comment.

t-8ch: Ack.

While looking at this, I found the following:
```
mmap(0x7fe08fa1b000, 31568, 0x3 /* PROT_READ|PROT_WRITE */, 0x32 /* MAP_PRIVATE|MAP_FIXED|MAP_ANONYMOUS */, -1, 0) = 0x7fe08fa1b000
```

In this, `PROT_READ` and `PROT_WRITE` have different colors.

src/kernel_time.h
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+#ifndef STRACE_KERNEL_TIME_H
+# define STRACE_KERNEL_TIME_H
+
+#define CLOCK_AUX			16
+#define MAX_AUX_CLOCKS			8
+#define CLOCK_AUX_LAST			(CLOCK_AUX + MAX_AUX_CLOCKS - 1)
+
+#endif /* !STRACE_KERNEL_TIME_H */
```
ldv-alt: Why can't we use <linux/time.h>?
If there are collisions with libc, could the code that needs these definitions be moved to a separate translation unit?
I'd like to avoid unnecessary duplication of upstream definitions if possible.

t-8ch: The definitions of `struct timespec` collide. (And some others, too IIRC).
It works on glibc due to some compatibility work, but it breaks on musl and probably others.

I'll try the new separate translation unit.

ldv-alt: I suppose e.g. `printclockname` could be easily moved to a separate translation unit.

t-8ch: That depends on where `CLOCKID_TO_FD()` is supposed to be coming from.
It looks like it is expected to come from libc, which would make it incompatible with `linux/time.h`.
However neither glibc nor musl define this macro.
The whole `CLOCKID_TO_FD()` block looks dead to me, and doesn't work on my distro build of strace.

So I would first fix that and then move it to a dedicated translation unit.

t-8ch: `printclockname.c` did not work out, because the functions used by the decoder all are declared in `src/defs.h` which needs libc `time.h`.
Instead I added `auxiliary_clock.c` only containing the `CLOCK_AUX` users.

esyr: > However neither glibc nor musl define this macro.

It's weird, as the kernel support is still there, and without any sings of it to be dropped.

esyr: Hm, as codesearch.debian.net suggests, `pipewire` and `linuxptp` just carry those definitions, also present in `clock_getres(2)`, around:
```
#define CLOCKFD 3
#define FD_TO_CLOCKID(fd)    ((~(clockid_t) (fd) << 3) | CLOCKFD)
#define CLOCKID_TO_FD(clk)	((unsigned int) ~((clk) >> 3))
```
I guess `strace` should do the same.

t-8ch: Agreed. I would have expected it to be part of *both* the UAPI and libc headers. But alas, it is not.
The macro once existed in the internal kernel headers and today is "documented" in the `clock_gettime()` manpage.

t-8ch: I also can't find any trace that the macro was ever provided in a way that would have been used by the current strace implementation. Given that the dynamic clock decoding doesn't even compile currently (see patch 1) I suspect this was dead code for a long time.

esyr: Okay, I'm starting to remember now, the macros were added to `include/linux/posix-timers.h` in `v2.6.39-rc1~508^2~11` "posix clocks: Introduce dynamic clocks", but that header has never been exposed as UAPI, as kernel developers do not tend to have a lot of care about exposing definitions for the user space, and an attempt to get it there[1] has failed.  Ultimately. the macros themselves have been removed from the kernel source in `v4.16-rc1~176^2~54` "posix-timers: Prevent UB from shifting negative signed value". So, yeah, it seems that the only way is to just bundle clockid-related definitions.

[1] https://lore.kernel.org/lkml/20190923130531.GA27774@asgard.redhat.com/

tests/printclockname.c
```diff
@@ -0,0 +1,97 @@
...
+	printf("clock_getres(CLOCK_THREAD_CPUTIME_ID, NULL) = 0\n");
+#endif
+
+	syscall(__NR_clock_getres, FD_TO_CLOCKID(0), NULL);
+#if XLAT_RAW
+	printf("clock_getres(-5, NULL)                  = -1 EINVAL (Invalid argument)\n");
+#elif XLAT_VERBOSE
+	printf("clock_getres(-5 /* FD_TO_CLOCKID(0) */, NULL) = -1 EINVAL (Invalid argument)\n");
+#else
+	printf("clock_getres(FD_TO_CLOCKID(0), NULL)    = -1 EINVAL (Invalid argument)\n");
```
esyr: Since the apparent reality is that `FD_TO_CLOCKID` is never normally provided, it should be printed as `-5 /* FD_TO_CLOCKID(0) */` for both abbreviated and verbose output.

t-8ch: Done.

t-8ch:
@esyr Could you take another look at this?

